PR #11, opened 2018-08-17
Re: Root Users

cerulean-skies:
Heya mcfarhat! I figured I would just give it a go :) Hope you're having a good one today. Hit me back up whenever. Weekend n all that, no rush at all.


mcfarhat:
Hey man,
Thanks for the work. I think its best if this is associated with the oncreate event, not the onresume? Give it a try there and see how the app responds.
Also you dont need to comment on every line of code, just wherever there could be lack of clarity, or to describe functions etc... feel free to check my comment style as well.
Let me know once you've double checked this.
Thanks!

mcfarhat:
Closing as stale — this PR dates to 2018, long predates the current codebase and the active development line, and is from an external (non-team) author. No longer applicable. Thanks regardless.

PR #30, opened 2025-07-01
Nur El Huda: Improved layout of gadgets on main screen

user12345-hub:
task 1(Improve layout of gadgets on main screen) line 4303 you can see the changes


mcfarhat:
change looks good

PR #73, opened 2026-07-23
AI Assistant chat popup + exercise-matching fix

AsmaaZeid:
## What this does
Adds an AI Assistant popup to the post-creation screen (`PostSteemitActivity`), letting users chat with Gemini (multi-turn) and insert the AI's reply directly into their post. Also includes a fix for exercise-name matching, and translations for the new AI strings.

## AI Assistant (post screen)
- Multi-turn chat popup using `gemini-flash-latest`, auth via `x-goog-api-key` header
- Automatic retry (up to 3x) on HTTP 503 "model overloaded"
- Graceful handling of blocked/unsafe responses (`ContentBlockedException`) instead of showing raw JSON
- New files: `ChatMessage.java`, `ChatAdapter.java`, associated chat bubble layouts
- UI matches existing toolbar button styling (`AppTheme.ToolbarIconButton`) and theme-adaptive popup background for light/dark mode

**Reference note:** Built with PR #52 as a starting reference, then diverged deliberately — this uses the existing `TextResponseCallback` interface and current Material Design styling rather than what #52 introduced, to stay consistent with the rest of the codebase.

## Exercise-matching fix
`Utils.findMatchingExercise()` previously only did substring containment, so AI-generated names (e.g. "Lat Pulldowns") often failed to match DB entries (e.g. "Full Range-Of-Motion Lat Pulldown"). Rewrote to try exact match first, then fall back to word-overlap scoring.

**Status: compiles cleanly, but not live-tested yet** — testing requires "Generate New Workout," which costs 300 AFIT tokens; I currently have 100. Flagging this so it isn't assumed to be verified end-to-end. Happy to test live once I can get the tokens, or if there's a way to test the function directly, let me know.

## Translations
Added first-pass translations for the 3 new AI strings (`ai_popup_title`, `ai_hint_input`, `ai_accept_button`) across all 11 existing locales. These are not native-speaker verified — flagging `yo` (Yoruba) in particular as the one I'm least confident in.

## Testing done
- AI popup tested manually on Pixel 8 emulator (API 34) — chat flow, insert-into-post, and retry/error handling all confirmed working


mcfarhat (changes requested):
Really solid first PR, @AsmaaZeid — welcome aboard! You got the security-sensitive parts right, which is the hard bit with an AI integration:

## 🟢 What's well done
- **API key handled correctly.** `GEMINI_API_KEY` comes from `BuildConfig` via `getLocalProperty('gemini.api.key')` reading `local.properties`, which is gitignored — so the key is never committed. You also moved it from the URL query string to the `x-goog-api-key` header (keys in URLs leak into logs), which is exactly right.
- **Thoughtful resilience.** 503 retry with backoff, and genuine content-safety handling — you check `promptFeedback.blockReason` and `finishReason` (SAFETY/PROHIBITED_CONTENT) and surface a friendly message instead of crashing. Nice.
- **The exercise-matching fix is a real improvement** — exact-match first, then word-overlap scoring with stopword removal and naive plural handling, plus a null-guard on `aiExerciseName`. Much better than the old substring `contains()` (which false-matched e.g. "press" against every "* press"). Multi-turn chat with proper `role` turns and full 14-locale translations are a nice touch too.

## 🔴 Required before merge
1. **Rebase on `develop`.** GitHub shows this as CONFLICTING/DIRTY, so it can't merge as-is:
   ```bash
   git switch feature/ai-post-screen
   git fetch origin
   git rebase origin/develop     # resolve conflicts, keeping develop's changes + your AI additions
   git push --force-with-lease origin feature/ai-post-screen
   ```

## 🟡 Please tidy (small)
2. **Remove the stale comment** next to the key: `// **REPLACE WITH YOUR ACTUAL API KEY!**`. The code already pulls from BuildConfig correctly — that comment invites someone to paste a literal key into source, which we don't want.
3. **Restrict the key.** Even via BuildConfig, the key ships inside the APK and can be extracted. Since this is shippable, please lock the Gemini key down in Google Cloud Console (restrict to this app's package + signing SHA-1, and restrict it to the Generative Language API). Longer term the robust pattern is to proxy the Gemini call through the actifit backend so the key never ships on-device — worth a follow-up card, not needed for this PR.
4. **Guard the async callbacks against a destroyed activity.** The AI/retry callbacks fire on the main looper and touch views; if the user leaves `PostSteemitActivity` mid-request the callback can hit a dead activity. A quick `if (isFinishing() || isDestroyed()) return;` (or null the callback in `onDestroy()`) avoids a rare crash.

Once you've rebased (and done 2), ping me and I'll re-review — items 3–4 can ride along. Genuinely good work for a first contribution. 👏

AsmaaZeid:
Hi Dr. Farhat, thank you so much for the detailed review, it really helped me understand the workflow better!

I went ahead and finished everything:
- Rebased onto develop and resolved the conflicts (turned out to be from the profile/companion strings work landing on develop after I branched)
- Removed the stale API key comment
- Added the isFinishing()/isDestroyed() guards to both onSuccess and onFailure in the AI callback

Item 3 (restricting the Gemini key in Google Cloud Console) I'll handle separately since you mentioned it's not blocking this PR.

Ready for another look whenever you have time. Thanks again for the guidance!

mcfarhat:
Re-reviewed, @AsmaaZeid — you nailed all the fixes from the last round. ✅

- **Stale API-key comment** — removed, the `API_KEY` line is clean now.
- **Destroyed-activity guards** — `if (isFinishing() || isDestroyed()) return;` is in place inside `runOnUiThread` in **both** `onSuccess` and `onFailure`, before any view access. Exactly the right spot — no more crash risk if the user leaves `PostSteemitActivity` mid-request.
- Your 14-locale AI strings check out too — keys consistent across all locales, apostrophes/ampersands properly escaped.

Nice, clean work. **One last thing before I can merge — a quick rebase.** Not your fault: your two translation PRs (#75 + #76) merged into `develop` today, and since this PR also touches all the `values-*/strings.xml` files, GitHub now shows it as conflicting again. The conflicts are purely additive (your AI strings vs. the DHF/menu strings that just landed — different keys, no real overlap), so resolving is just 'keep both':

```bash
git switch feature/ai-post-screen
git fetch origin
git rebase origin/develop
# For each conflicted values-*/strings.xml: KEEP BOTH sides —
# your AI keys AND the DHF/menu keys from develop (they don't overlap).
git add app/src/main/res/values*/strings.xml
git rebase --continue
git push --force-with-lease origin feature/ai-post-screen
```

Once it's green, ping me and I'll merge it straight away. This was the last gate — solid feature. 👏

AsmaaZeid:
Rebase done and pushed! Sorry for the delay — had pretty unstable internet the past couple of days, which made this take longer than it should have.

A couple of the locale files (values-de, values-yo) had genuine duplicate/malformed XML issues from the merge that I caught and fixed before pushing — confirmed with a clean gradlew assembleDebug build. Ready for the merge whenever you get a chance!

mcfarhat:
@AsmaaZeid thanks — the rebase is done and the feature code itself is in good shape (the destroyed-activity lifecycle guards and the stale API-key comment are both handled, and the exercise-matching logic looks solid). But the branch as pushed does **not** build — two issues slipped in during the locale-file conflict resolution, so `./gradlew assembleDebug` fails at resource compilation:

1. **`app/src/main/res/values-yo/strings.xml`** — malformed XML: there are two `</resources>` closing tags (a stray second one near the end, after a blank line). aapt2 rejects it. Remove the extra tag so the file ends with exactly one `</resources>`.
2. **`app/src/main/res/values-de/strings.xml`** — duplicate resource key: `profile_metrics_legend` is defined twice (around line 742 and line 760). Keep the intended version (the one with 'km') and delete the other.

Minor: `values-uk/strings.xml` has a ~990-line whitespace-only reindent that's pure diff noise — please revert the reformat to keep the diff reviewable.

Please run `./gradlew assembleDebug` locally and confirm it's green before pushing and re-pinging — that catches exactly this class of issue (there's no CI on the branch to catch it for us). Once it builds clean, we're ready to merge.

AsmaaZeid:
Fixed all three:
- values-de: removed the duplicate profile_metrics_legend, kept the one with "km"
- values-yo: removed the stray extra </resources>
- values-uk: reverted the accidental ~484-line reindent, so the diff against develop is now just the 3 new AI strings

Confirmed clean via `./gradlew assembleDebug` — BUILD SUCCESSFUL. Pushed. Ready whenever you get a chance!

PR #72, opened 2026-07-21
Implement signup wizard

Adam-Diab1:
This PR implements the signup wizard.

- Added a multi-step signup wizard
- Added secure password generation using SecureRandom
- Added on-chain username availability check
- Integrated the actifit.signup payment flow

PR #81, opened 2026-07-29
Implement signup wizard

Adam-Diab1:
This PR implements the signup wizard.

- Added a multi-step signup wizard
- Added secure password generation using SecureRandom
- Added on-chain username availability check
- Integrated the actifit.signup payment flow

"This PR replaces the accidentally closed PR #72 and includes the requested review fixes."


mcfarhat (changes requested):
Great turnaround, @Adam-Diab1 — the recut is clean and the critical items from #72 are genuinely fixed:
- ✅ No secrets file in history (0 commits touch `unofficial_strings.xml`), `.example` added.
- ✅ SecureRandom password generation; payment poll now has `MAX_POLL_ATTEMPTS` (60 → ~5 min) + timeout + `removeCallbacksAndMessages` cleanup + a cancelable dialog; strings externalized to `signup_strings.xml`; `SignupActivity` registered (`exported="false"`); and nothing logs the password to logcat. Nicely done.

A few things to fix before this can merge:

**🔴 1. Manifest regression (blocker).** Your `AndroidManifest.xml` drops two permissions that `develop` has — `android.permission.health.READ_HEALTH_DATA` and `android.permission.health.WRITE_HEALTH_DATA` (develop has 21 active `uses-permission`, this PR has 19). That would break health-data access. It's caused by a full ~335-line manifest reformat. **Fix:** re-cut `AndroidManifest.xml` from develop's current version and add **only** the `<activity android:name=".SignupActivity" .../>` block — don't reorder/reformat the rest. The diff should then be a few lines, not 335.

**🔴 2. Wrong verify endpoint (correctness).** You POST to `https://actifit.io/api/signup_verify`, which isn't a real API endpoint — verification will fail against production. The correct one is `POST https://api.actifit.io/confirmPayment` (see the Swagger at api.actifit.io/api-docs). Good news: your payload fields already match its body (`new_pass`, `promo_code`, `new_account`, `memo`, `usd_invest`, `referrer`, `email`), so it's essentially just the URL — please confirm against the swagger and switch to it.

**🟠 3. Drop the committed `.idea/` files** (`deploymentTargetSelector.xml`, `gradle.xml`, `misc.xml`) and gitignore the `.idea/` folder.

**🟠 4. Trim unrelated build churn** — `build.gradle`, `settings.gradle`, `gradle.properties`, `gradle/wrapper/gradle-wrapper.properties`, `gma_ad_services_config.xml`. Unless a change there is genuinely required for signup, revert them so the PR stays scoped to the wizard.

**Note (no action needed):** keep sending `new_pass` exactly as-is — it matches the current `/confirmPayment` contract, and we'll handle moving to a passwordless (public-keys-only) signup separately as a platform-level change. Don't change that in this PR.

Once the manifest is scoped down (permissions restored) and the endpoint points at `api.actifit.io/confirmPayment`, ping me and I'll re-review. The wizard itself is solid work. 👍

mcfarhat (changes requested):
Good work, @Adam-Diab1 — the two blockers from last round are genuinely fixed:
- ✅ **Manifest regression resolved** — active permissions are back to 21 (matching develop), `READ_HEALTH_DATA` / `WRITE_HEALTH_DATA` restored, nothing dropped.
- ✅ **Endpoint corrected** — now `https://api.actifit.io/confirmPayment` (right host + real endpoint).

**One thing still to fix before merge — the `.idea/` cleanup (3rd time on this, so let's close it for good):** instead of removing the committed IDE files, this PR now commits the **entire `.idea/` folder (~14 files** — `compiler.xml`, `misc.xml`, `deploymentTargetSelector.xml`, `deviceManager.xml`, `gradle.xml`, `migrations.xml`, etc.). IDE project files must never be committed to a shared repo — they cause churn and conflicts for every dev. Fix:

```bash
printf '\n.idea/\n' >> .gitignore
git rm -r --cached .idea
git commit -m "chore: stop tracking .idea IDE files"
git push
```

Also drop **`app/build.gradle`** from the PR unless a change there is genuinely required for signup.

That's the only thing between this and merge — the signup wizard itself and the two real fixes are solid. Ping me once `.idea/` is gone and I'll merge it. 👍

Adam-Diab1:
Good Evening Dr  Farhat, is all good now?
________________________________

mcfarhat:
⚠️ **This merge has been reverted from `develop`** in commit `6cd8cc3`.

The signup wizard does not conform to the live `api.actifit.io/confirmPayment` contract and cannot complete a real signup:
- auth token sent under `sec_param` instead of the required `confirm_payment_token` field → server returns `{}`, wizard times out (blocks promo **and** paid paths);
- no USD→HIVE conversion and `steem_invest` never sent → on-chain amount check compares against `NaN`;
- currency hardcoded to `HIVE` while the UI offers HBD.

Rework tracked in #85 (rebuilt on current `develop`). Detailed feedback also provided to the developer on Trello.

PR #82, opened 2026-07-30
Improve HC dashboard readability + distance/calories; badge fix; 0-kcal fix

mcfarhat:
Polishes the multi-ring Health Connect dashboard after a moderator reported it was hard to read. Verified on-device (light + dark).

## Changes

**Dashboard readability** (`AuraView`, `activity_main.xml`, `MainActivity`)
- Opt-in opaque **center disc** in the card color so the counter sits on a clean surface (not over the arcs/glow). Disc is off by default (`AuraView.setCenterFillColor`, `TRANSPARENT`) — only the dashboard opts in.
- **Dark mode:** ring tracks now use a neutral light groove (the old color-at-21%-alpha vanished on the dark card) and dark ring colors are lifted to a minimum brightness so no ring is lost. Light mode unchanged.
- Step count 32→26sp, animal 44→34dp so content fits the clear center; labels → `md_theme_textSecondary`.

**Distance + calories** (`MainActivity`)
- Extra HC metrics shown under the step count as `📏 distance  🔥 calories` (**bold**), distance honoring the user's measurement system via `Utils.formatDistance`. Estimated values (no real HC source) are prefixed `≈`. `textDirection=ltr` keeps order stable in RTL locales; `kcal` is a string resource.

**Fix "0 kcal"** (`HealthConnectManager`)
- `readTodayMetrics` coerced a missing distance/calorie aggregate to `0` instead of `-1`, so "granted but no data source" showed `0 kcal` instead of the intended step-derived estimate. Now returns `-1` → caller estimates. (This also silently corrects `ProfileActivity`'s calorie/distance estimate, which already handled `-1`.) Caveat: a genuine zero is now indistinguishable from "no source" and shows as an `≈` estimate.

**Notification badge** (`activity_main.xml`, `MainActivity`, `notif_badge_pill`)
- Was a fixed 18dp oval that clipped 3-digit counts. Now a rounded "stadium": circle for 1–2 digits, pill for larger, with the count centered (dropped the `<sup>` that pushed it to the top).

## Review follow-ups (commit 557eb27)
Addresses the code review:
1. Step-count color: XML `onSurface` never rendered (overwritten by `ChartManager`); real issue was dark-green `#008000` ≈2.4:1 on dark → brightened night `actifitDarkGreen` to `#4CAF50`, reverted the dead XML.
2. `tv_step_pct_hc` two-writer conflict → `ChartManager` passes `null` for `tvPct` in HC multi-ring; metrics own it (no longer vanish on refresh).
3. Center disc leaking into `ProfileActivity` → disc is now opt-in.
4. Metrics kept in center, bold; RTL-safe; `kcal` localized.
5. Minors: fixed the same `<sup>` bug in the unused `ApiManager` duplicate; shared `@color/badge_green`; reuse `Utils.isDarkModeActive`; night computed once/onDraw + cached HSV; dropped redundant ternary.

## Testing
- `compileReleaseJavaWithJavac` + `assembleRelease` pass. Installed and verified on-device across light/dark.

### DIFF
--- a/app/src/main/java/io/actifit/fitnesstracker/actifitfitnesstracker/ApiManager.java
+++ b/app/src/main/java/io/actifit/fitnesstracker/actifitfitnesstracker/ApiManager.java
@@ -570,7 +570,7 @@ public class ApiManager {
         JsonArrayRequest transactionRequest = new JsonArrayRequest(Request.Method.GET, notificationsUrl, null, notificationsListArray -> {
             if (notificationsListArray != null && notificationsListArray.length() > 0) {
                 String count = notificationsListArray.length() < 1000 ? notificationsListArray.length() + "" : "999+";
-                notifCount.setText(Html.fromHtml("<sup><small>" + count + "</small></sup>"));
+                notifCount.setText(count); // plain centered text (badge background handles the shape)
                 notifCount.setVisibility(View.VISIBLE);
             }
         }, error -> {});
--- a/app/src/main/java/io/actifit/fitnesstracker/actifitfitnesstracker/AuraView.java
+++ b/app/src/main/java/io/actifit/fitnesstracker/actifitfitnesstracker/AuraView.java
@@ -136,6 +136,9 @@ public class AuraView extends View {
     private final Paint arcPaint = new Paint(Paint.ANTI_ALIAS_FLAG);
     private final Paint glowPaint = new Paint(Paint.ANTI_ALIAS_FLAG);
     private final Paint emojiPaint = new Paint(Paint.ANTI_ALIAS_FLAG);
+    // opaque disc painted in the card colour behind the centre content of the multi-ring dashboard,
+    // so the step count / labels sit on a clean, high-contrast surface instead of over the arcs+glow
+    private final Paint centerFillPaint = new Paint(Paint.ANTI_ALIAS_FLAG);
     private final RectF arcBounds = new RectF();
 
     // secondary activity-ring colours (outer ring uses the animal's signature colour)
@@ -235,6 +238,25 @@ public class AuraView extends View {
         return Color.HSVToColor(hsv);
     }
 
+    private boolean nightMode() {
+        return (getResources().getConfiguration().uiMode
+                & android.content.res.Configuration.UI_MODE_NIGHT_MASK)
+                == android.content.res.Configuration.UI_MODE_NIGHT_YES;
+    }
+
+    /**
+     * In dark mode, lifts a ring colour to a minimum brightness so dark companion colours
+     * (e.g. Penguin slate) don't vanish against the dark card. No-op in light mode / for
+     * already-bright colours.
+     */
+    private int visibleRingColor(int color) {
+        if (!nightMode()) return color;
+        float[] hsv = new float[3];
+        Color.colorToHSV(color, hsv);
+        if (hsv[2] < 0.72f) hsv[2] = 0.72f;
+        return Color.HSVToColor(hsv);
+    }
+
     @Override
     protected void onDraw(Canvas canvas) {
         super.onDraw(canvas);
@@ -281,6 +303,16 @@ public class AuraView extends View {
             }
             drawRing(canvas, cx, cy, outer, stroke, color, fillFraction, true, glowRadius, intensity, breath);
 
+            // clean opaque centre in the card colour: fills the clear zone inside the innermost ring
+            // (and masks any inward glow bleed) so overlaid counter text reads crisply
+            float discRadius = inner - stroke * 0.5f;
+            if (discRadius > 0) {
+                centerFillPaint.setStyle(Paint.Style.FILL);
+                centerFillPaint.setColor(getContext().getColor(
+                        io.actifit.fitnesstracker.actifitfitnesstracker.R.color.md_theme_cardBackground));
+                canvas.drawCircle(cx, cy, discRadius, centerFillPaint);
+            }
+
             if (showAnimal) {
                 drawAnimal(canvas, cx, cy, outer, R);
             }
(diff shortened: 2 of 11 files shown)


mcfarhat:
## Code review notes

Reviewed the branch against the surrounding code (`ChartManager`, `TrackingManager`, `ProfileActivity`, layouts). The `HealthConnectManager` `0` → `-1` fix is solid — it also silently repairs a second consumer (see "Good" below). Three of the UI changes, though, look like they won't behave as the description claims. Items 1–3 are worth validating on-device before merge.

---

### 1. The step-count colour change never renders — it's overwritten at runtime

`activity_main.xml` sets `tv_step_count_hc` to `@color/md_theme_onSurface`, but `ChartManager.updateRing()` overrides it on every HC refresh:

```java
// ChartManager.java:137
tvCount.setTextColor(color);   // color = actifitRed, or actifitDarkGreen past the milestone
```

`TrackingManager.java:288` calls `chartManager.displayActivityChartHealthConnect(...)` immediately **before** the async `readTodayMetrics(...)` → `updateHcDashboardRings(...)`, so the runtime colour always wins over the XML.

Net effect in dark mode for a user past the step milestone: `actifitDarkGreen` `#008000` on card `#1E1E1E` ≈ **2.4:1 contrast** — worse than what this PR set out to fix.

**Suggested fix:** move the contrast change into `updateRing()` (or stop it setting `tvCount`'s colour in HC multi-ring mode) rather than in XML.

**To validate:** open the dashboard in HC mode in dark theme with > milestone steps and confirm whether the counter is green/red rather than white.

---

### 2. `tv_step_pct_hc` now has two writers that fight

Two code paths write the same TextView:

- `ChartManager.updateRing()` (line 140) → `"N% to goal"`
- `MainActivity.updateHcDashboardRings()` (line 2852) → `"📏 …  🔥 …"`

On the `TrackingManager` sync path the metrics line happens to land last, so it looks correct. But `MainActivity.java:2538` (`displayActivityChartHealthConnect`) also refreshes the ring **without** re-reading HC metrics — that path reverts the label back to `"N% to goal"` and the distance/calorie line disappears until the next full sync.

**Suggested fix:** give the metrics their own TextView, or have `updateRing()` skip `tvPct` when the multi-ring HC view is active.

**To validate:** leave the dashboard open and trigger a step refresh that goes through `MainActivity.displayActivityChartHealthConnect` (without a full HC metrics read) — the km/kcal line should flip back to "% to goal".

---

### 3. The centre disc leaks into ProfileActivity (out of stated scope)

The opaque disc is drawn unconditionally in the `multiRing` branch (`AuraView.java:306-314`), and `ProfileActivity.java:185` also uses `setActivityRings(...)`.

That AuraView sits on `@color/md_theme_background` (`#F5F5F5` light / `#121212` dark), **not** `@color/md_theme_cardBackground` (`#FFFFFF` / `#1E1E1E`). So a hard-edged ~99dp circle in the wrong shade now appears behind the profile companion hero, in both themes.

**Suggested fix:** make the disc opt-in / configurable rather than a hardcoded resource lookup inside `onDraw` — e.g. `setCenterFillColor(int color)` defaulting to transparent, set only from `updateHcDashboardRings`.

**To validate:** open the Profile screen (light and dark) and look for a lighter/darker circle behind the animated companion.

---

### 4. The centre content likely still overflows the "clear" disc

At the dashboard's 184dp AuraView: `R = 92dp` → `stroke ≈ 7.8`, `inner ≈ 38.3`, so `discRadius ≈ 34.4dp` — a **~69dp clear circle**.

The centred column is animal 34dp + 26sp + 11sp + 10sp ≈ **90dp tall**, and the new metrics row (`📏 12.34 km    🔥 456 kcal`, ~24 chars at 10sp ≈ 110dp wide) sits roughly 40dp below centre and ~55dp out — putting its ends ~68dp from centre, i.e. back on the middle/outer rings.

Shrinking the companion 44 → 34dp bought ~10dp of headroom; the added third line spends it. Also note the text is `sp` while the ring geometry is `dp`, so this degrades further at large accessibility font scales.

**To validate:** device font size at Largest, with a long value (e.g. `12.34 km` / `1234 kcal`), and check the km/kcal row against the rings.

---

### Minor

- **PR description is stale.** Two commits landed after the body was written (`ea4abb7`, `57df531` — notification badge pill, centring, and the `Html.fromHtml("<sup>…")` removal at `MainActivity.java:4265`). None of that is mentioned in the title or body.
- `ApiManager.loadNotifCount()` (line 573) is an **unused duplicate** of the badge code that still carries the `<sup><small>` wrap. Fix it alongside, or delete it — otherwise the bug returns if it's ever wired up.
- `notif_badge_pill.xml` hardcodes `#0a660a` with a "matches circle_badge" comment instead of extracting a shared colour resource. `circle_badge` is still used in `post_entry.xml`, so the two will drift.
- `"kcal"` and the emoji are hardcoded, unlocalized strings — the app ships 12 locales, and Arabic is RTL where `📏 12.34 km    🔥 456 kcal` will bidi-reorder unpredictably.
- `AuraView.nightMode()` duplicates the existing `Utils.isDarkModeActive(Context)` helper (`Utils.java:1709`).
- Per-frame cost in an `invalidate()`-driven view: a `getColor()` resource lookup, 3× `getResources().getConfiguration()` (once per `drawRing`), and a `new float[3]` per ring per frame. Worth caching on attach / `onConfigurationChanged`.
- `night ? visibleRingColor(color) : color` (`AuraView.java:344`) is redundant — `visibleRingColor` already no-ops in light mode.

---

### Good

The `HealthConnectManager` `0` → `-1` fix is correct, and it also silently corrects `ProfileActivity.realHcDistanceKm()` / `realHcKcal()` (lines 199-205), which already treated `-1` as "derive" but were being fed a fake `0` via the `hcTodayDistanceM` / `hcTodayKcal` prefs. Worth calling out in the PR body.

One caveat worth documenting: a genuine zero (permission granted, source present, no records today) is indistinguishable from "no source", so a real `0` will now display as an `≈` estimate.

mcfarhat:
Thanks for the thorough review — all four substantive points were spot on. Addressed in 557eb27:

**1. Step-count contrast never rendered.** Correct — `ChartManager.updateRing` overwrites the XML color. The real defect was dark-green `#008000` ≈ 2.4:1 on the dark card, so I brightened the *night* `actifitDarkGreen` → `#4CAF50` (fixes it app-wide, milestone coloring preserved) and reverted the dead `onSurface` XML to `actifitRed`.

**2. Two writers on `tv_step_pct_hc`.** Fixed by passing `null` for `tvPct` from `displayActivityChartHealthConnect`, so `updateRing` no longer writes "% to goal" over the metrics line — it no longer flips back on the `displayActivityChartHealthConnect` refresh path.

**3. Disc leaked into Profile.** Made it opt-in: `AuraView.setCenterFillColor(int)` (default `TRANSPARENT`), set only from `updateHcDashboardRings` to the card color. Profile's rings never opt in → disc-free.

**4. Center overflow.** Per product call the metrics stay in the center but are now **bold** for legibility; `textDirection=ltr` keeps the emoji/number order stable in RTL, and `kcal` is now a string resource. (Long-value + max-font-scale overflow is a known edge; can revisit if it bites.)

**Minors:** fixed the same `<sup>` bug in the unused `ApiManager.loadNotifCount` duplicate; extracted `@color/badge_green` shared by `circle_badge` + `notif_badge_pill`; swapped `AuraView.nightMode()` for `Utils.isDarkModeActive`; night mode is now computed once per `onDraw` with a cached HSV array (drops the per-ring config lookups + allocation); removed the redundant light-mode ternary. PR description refreshed.

Good call on the 0→-1 fix also repairing `ProfileActivity`'s estimate — and yes, a genuine zero now shows as an `≈` estimate; noted in the PR body.

PR #83, opened 2026-07-30
Unify multi-ring dashboard across all tracking modes + Fitbit distance/calories

mcfarhat:
Brings the multi-metric ring dashboard (steps/distance/calories + the km·cal legend) to **every** tracking mode. Previously only Health Connect had it; sensors and Fitbit modes still showed the old single ring with no km/cal.

## Changes

**Generalized render** (`MainActivity`)
- `updateHcDashboardRings` → `renderDashboardRings(auraId, materialId, metricsTvId, steps, distanceMeters, kcal)`, with per-mode wrappers: `updateHcDashboardRings`, `updateDeviceDashboardRings` (estimates), `updateFitbitDashboardRings`.
- `distanceMeters/kcal < 0` ⇒ no real source ⇒ step-derived estimate with a `≈` prefix (per-metric).

**Layout** (`activity_main.xml`)
- Added the multi-ring `AuraView` + bold metrics line to the **sensors** and **Fitbit** blocks (same center disc / contrast / sizing as HC). The old `CircularProgressIndicator` is hidden once the rings render.

**Wiring** (`ChartManager`)
- `displayActivityChart` / `displayActivityChartFitbit` now pass `null` for `tvPct`, so the metrics line (owned by MainActivity) isn't overwritten with "% to goal".

**Fitbit real distance + calories** (`MainActivity`)
- The sync now also pulls the tracker's `distance` and `calories` daily time series (`sumFitbitTrackerMetric`), converts distance to meters using the account's `distanceUnit` (km vs miles), caches them (`fitbitDistanceM`, `fitbitCalories`), and feeds the Fitbit rings — **real** values, not estimates.

## Per-mode behavior
| Mode | Distance | Calories |
|---|---|---|
| Health Connect | real (HC) | real if source, else `≈` estimate |
| **Fitbit** | **real (Fitbit)** | **real (Fitbit)** |
| **Sensors** | `≈` estimate | `≈` estimate |

## Testing
- `compileReleaseJavaWithJavac` + `assembleRelease` pass.
- Verified **HC mode** on-device through the new generalized path (real `1.56 km`, `≈113 kcal`).
- ⚠️ **Sensors** and **Fitbit** modes need on-device verification (couldn't switch tracking mode via adb on a release build). Fitbit distance/calorie parsing + the `distanceUnit` conversion should be sanity-checked against a real Fitbit-connected account.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/app/src/main/java/io/actifit/fitnesstracker/actifitfitnesstracker/ChartManager.java
+++ b/app/src/main/java/io/actifit/fitnesstracker/actifitfitnesstracker/ChartManager.java
@@ -145,10 +145,10 @@ public class ChartManager {
             stepRing = ((android.app.Activity) context).findViewById(R.id.step_ring);
             android.widget.TextView tvCount = ((android.app.Activity) context).findViewById(R.id.tv_step_count);
             android.widget.TextView tvGoal = ((android.app.Activity) context).findViewById(R.id.tv_step_goal);
-            android.widget.TextView tvPct = ((android.app.Activity) context).findViewById(R.id.tv_step_pct);
             if (stepRing == null) return;
 
-            updateRing(stepRing, tvCount, tvGoal, tvPct, stepCount, animate);
+            // tv_step_pct owned by MainActivity.updateDeviceDashboardRings (distance/calorie metrics)
+            updateRing(stepRing, tvCount, tvGoal, null, stepCount, animate);
 
             if (stepCount > 2000) {
                 if (BtnWaves != null && (BtnWaves.getAnimation() == null || !BtnWaves.getAnimation().hasStarted())) {
@@ -174,10 +174,10 @@ public class ChartManager {
             stepRingFitbit = ((android.app.Activity) context).findViewById(R.id.step_ring_fitbit);
             android.widget.TextView tvCount = ((android.app.Activity) context).findViewById(R.id.tv_step_count_fitbit);
             android.widget.TextView tvGoal = ((android.app.Activity) context).findViewById(R.id.tv_step_goal_fitbit);
-            android.widget.TextView tvPct = ((android.app.Activity) context).findViewById(R.id.tv_step_pct_fitbit);
             if (stepRingFitbit == null) return;
 
-            updateRing(stepRingFitbit, tvCount, tvGoal, tvPct, stepCount, animate);
+            // tv_step_pct_fitbit owned by MainActivity.updateFitbitDashboardRings (distance/calorie metrics)
+            updateRing(stepRingFitbit, tvCount, tvGoal, null, stepCount, animate);
 
             if (stepCount > 2000) {
                 if (BtnWaves != null && (BtnWaves.getAnimation() == null || !BtnWaves.getAnimation().hasStarted())) {
--- a/app/src/main/java/io/actifit/fitnesstracker/actifitfitnesstracker/ChartManager.java
+++ b/app/src/main/java/io/actifit/fitnesstracker/actifitfitnesstracker/ChartManager.java
@@ -149,6 +149,9 @@ public class ChartManager {
 
             // tv_step_pct owned by MainActivity.updateDeviceDashboardRings (distance/calorie metrics)
             updateRing(stepRing, tvCount, tvGoal, null, stepCount, animate);
+            // render the multi-ring dashboard + metrics here (not in MainActivity's wrapper) so every
+            // caller is covered — including TrackingManager.useDefaultTrackingMethod's direct calls
+            if (context instanceof MainActivity) ((MainActivity) context).updateDeviceDashboardRings(stepCount);
 
             if (stepCount > 2000) {
                 if (BtnWaves != null && (BtnWaves.getAnimation() == null || !BtnWaves.getAnimation().hasStarted())) {
@@ -178,6 +181,8 @@ public class ChartManager {
 
             // tv_step_pct_fitbit owned by MainActivity.updateFitbitDashboardRings (distance/calorie metrics)
             updateRing(stepRingFitbit, tvCount, tvGoal, null, stepCount, animate);
+            // render rings + metrics here so TrackingManager's direct calls are covered too
+            if (context instanceof MainActivity) ((MainActivity) context).updateFitbitDashboardRings(stepCount);
 
             if (stepCount > 2000) {
                 if (BtnWaves != null && (BtnWaves.getAnimation() == null || !BtnWaves.getAnimation().hasStarted())) {
--- a/app/src/main/java/io/actifit/fitnesstracker/actifitfitnesstracker/MainActivity.java
+++ b/app/src/main/java/io/actifit/fitnesstracker/actifitfitnesstracker/MainActivity.java
@@ -1140,33 +1140,38 @@ public class MainActivity extends BaseActivity {
                                 trackedActivityCount += parseInt(stepActivityArray.getJSONObject(i).getString("value"));
                             }
 
-                            displayActivityChartFitbit(trackedActivityCount, true);
-                            Calendar mCalendar = Calendar.getInstance();
+                            // also fetch Fitbit's own distance + activity calories for the multi-metric
+                            // rings. activityCalories (not tracker/calories, which includes BMR) keeps
+                            // parity with Health Connect's active-calories. Distance is returned in the
+                            // account's unit, so convert to metres via the profile's distanceUnit.
+                            float fbDistRaw = sumFitbitActivityResource(fitbit, "tracker/distance", targetDate);
+                            float fbCal = sumFitbitActivityResource(fitbit, "activityCalories", targetDate);
+                            float fbDistM = -1f;
+                            if (fbDistRaw >= 0) {
+                                String distUnit = fitbit.getFieldFromProfile("distanceUnit"); // "METRIC"/"en_US" or a sentinel on failure
+                                String u = (distUnit == null) ? "" : distUnit.toUpperCase(java.util.Locale.ROOT);
+                                if (u.contains("US")) {
+                                    fbDistM = fbDistRaw * 1609.344f;      // imperial account -> miles
+                                } else if (u.contains("METRIC")) {
+                                    fbDistM = fbDistRaw * 1000f;          // metric account -> km
+                                }
+                                // else: unknown/failed profile lookup -> leave -1 so distance degrades
+                                // to a "≈" estimate rather than guessing the wrong unit
+                            }
 
-                            editor.putString("fitbitLastSyncDate",
-                                    new SimpleDateFormat("yyyyMMdd").format(
-                                            mCalendar.getTime()));
+                            Calendar mCalendar = Calendar.getInstance();
+                            String todayStamp = new SimpleDateFormat("yyyyMMdd").format(mCalendar.getTime());
+                            editor.putString("fitbitLastSyncDate", todayStamp);
                             editor.putLong("fitbitLastSyncTime", System.currentTimeMillis());
                             // TODO: demo data, replace when go live
                             editor.putInt("fitbitSyncCount", trackedActivityCount);// 6543);//
-                            editor.apply();
-
-                            // also fetch Fitbit's own distance + calories for the multi-metric rings.
-                            // Distance comes back in the account's unit (km/mi), so convert to metres.
-                            float fbDistRaw = sumFitbitTrackerMetric(fitbit, "distance", targetDate);
-                            float fbCal = sumFitbitTrackerMetric(fitbit, "calories", targetDate);
-                            float fbDistM = -1f;
-                            if (fbDistRaw >= 0) {
-                                String distUnit = fitbit.getFieldFromProfile("distanceUnit");
-                                boolean miles = distUnit != null && distUnit.toUpperCase().contains("US");
-                                fbDistM = fbDistRaw * (miles ? 1609.344f : 1000f);
-                            }
                             editor.putFloat("fitbitDistanceM", fbDistM);
                             editor.putFloat("fitbitCalories", fbCal);
+                            editor.putString("fitbitMetricsDate", todayStamp); // freshness stamp: stale -> estimate
                             editor.apply();
-                            final int fbSteps = trackedActivityCount;
-                            final float fbDistFinal = fbDistM, fbCalFinal = fbCal;
-                            runOnUiThread(() -> updateFitbitDashboardRings(fbSteps, fbDistFinal, fbCalFinal));
+
+                            // render after the metrics are cached, so the rings pick up fresh values
+                            displayActivityChartFitbit(trackedActivityCount, true);
                         } else {
                             Log.d(MainActivity.TAG, "No auto-tracked activity found for today");
                         }
@@ -2546,14 +2551,16 @@ public class MainActivity extends BaseActivity {
     }
 
     /**
-     * Sums a Fitbit daily tracker time-series resource ("distance" / "calories"). Returns the raw
-     * summed value (distance in the account's unit, calories in kcal), or -1 if unavailable so
-     * callers fall back to a step-derived estimate. Runs on a background thread.
+     * Sums a Fitbit daily activity time-series resource by its path under "activities/"
+     * (e.g. "tracker/distance", "activityCalories"). Returns the raw summed value (distance in the
+     * account's unit, calories in kcal), or -1 if unavailable so callers fall back to a step-derived
+     * estimate. Note the deliberate asymmetry: an empty/unavailable dataset returns -1 (→ estimate),
+     * whereas a dataset of genuine zeros sums to 0 (shown as a real "0"). Runs on a background thread.
      */
-    private float sumFitbitTrackerMetric(NxFitbitHelper fitbit, String metric, String targetDate) {
+    private float sumFitbitActivityResource(NxFitbitHelper fitbit, String resourcePath, String targetDate) {
         try {
-            JSONObject list = fitbit.getActivityByDate(metric, targetDate);
-            JSONArray arr = list.getJSONArray("activities-tracker-" + metric);
+            JSONObject list = fitbit.getActivityResource(resourcePath, targetDate);
+            JSONArray arr = list.getJSONArray("activities-" + resourcePath.replace("/", "-"));
             if (arr.length() == 0) return -1f;
             float sum = 0f;
             for (int i = 0; i < arr.length(); i++) {
@@ -2561,19 +2568,14 @@ public class MainActivity extends BaseActivity {
             }
             return sum;
         } catch (Exception e) {
-            Log.e(TAG, "Fitbit " + metric + " fetch failed: " + e.getMessage());
+            Log.e(TAG, "Fitbit " + resourcePath + " fetch failed: " + e.getMessage());
             return -1f;
         }
     }
 
     private void displayActivityChartFitbit(final int stepCount, final boolean animate) {
         currentDisplayedStepCount = stepCount;
-        chartManager.displayActivityChartFitbit(stepCount, animate);
-        // multi-ring + metrics using Fitbit's own distance/calorie readings when synced (else estimate)
-        SharedPreferences fbPrefs = getSharedPreferences("actifitSets", MODE_PRIVATE);
-        float fbDist = fbPrefs.getFloat("fitbitDistanceM", -1f);
-        float fbCal = fbPrefs.getFloat("fitbitCalories", -1f);
-        updateFitbitDashboardRings(stepCount, fbDist, fbCal);
+        chartManager.displayActivityChartFitbit(stepCount, animate); // also renders the Fitbit rings + metrics
         updateNudgeCard(stepCount);
         if (animate) checkMilestoneCelebration(stepCount);
     }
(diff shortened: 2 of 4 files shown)


mcfarhat:
## Code review notes

Reviewed at `4572d53` against `develop` (which already has the squashed #82). `compileDebugJavaWithJavac` passes — only the pre-existing Java 8 `source/target value 8 is obsolete` warnings.

The refactor itself is clean: `renderDashboardRings` is the right generalization, all three ring blocks sit inside the same `md_theme_cardBackground` card (layout lines 431–1000) so the opt-in centre disc colour matches in every mode, and the sensors/Fitbit layouts correctly mirror the HC treatment (34dp companion, 26sp count, `textSecondary` labels, bold + `textDirection=ltr` metrics line).

Two blockers though, and both sit in the parts the description already flags as unverified.

---

## Blocking

### 1. Fitbit distance/calories are written on exactly one occasion, then rendered forever as "real"

`fitbitDistanceM` / `fitbitCalories` are only ever written at `MainActivity.java:1164-1165` — inside the `if (returnUrl != null)` OAuth-callback block in `onCreate`. (`fitbitSyncCount` has the same single writer at :1151, so this staleness is pre-existing for steps.)

Every other path — normal launch, `TrackingManager.useDefaultTrackingMethod`, the mode-switch buttons, `onPostExecute` — reads the cached prefs in `displayActivityChartFitbit` (:2574-2575) and renders them with **no `≈` prefix and no freshness check**:

```java
float fbDist = fbPrefs.getFloat("fitbitDistanceM", -1f);
float fbCal  = fbPrefs.getFloat("fitbitCalories", -1f);
updateFitbitDashboardRings(stepCount, fbDist, fbCal);
```

So a value captured on the day the user connected Fitbit is displayed as today's *real* distance/calories, indefinitely. The per-mode table in the description promises "Fitbit → real / real", which makes the stale number look authoritative rather than estimated.

The HC path already solved exactly this with `hcMetricsDate` + `hcMetricsFresh()` (see `ProfileActivity.hcMetricsFresh`); the Fitbit cache has no date stamp at all.

**Suggested fix:** stamp a `fitbitMetricsDate` alongside the values and coerce anything not from today back to `-1` (→ `≈` estimate). Better still, move the metric fetch to whatever runs on ordinary Fitbit syncs rather than only the OAuth redirect.

---

### 2. Five call sites bypass the new wrappers → blank label and no rings

The `tvPct → null` change is in `ChartManager` (shared by every caller), but the replacement text is only written by MainActivity's wrappers. `TrackingManager` calls ChartManager **directly**, bypassing them:

- `TrackingManager.java:374` — Fitbit
- `TrackingManager.java:424, 435, 437, 448` — sensors

All five are inside `useDefaultTrackingMethod()`, which has 8 call sites and is the startup / fallback path.

Through those paths:
- `dashboard_rings` / `dashboard_rings_fitbit` stay `visibility="gone"` — only `renderDashboardRings` ever makes them VISIBLE, so the rings never appear
- `tv_step_pct` / `tv_step_pct_fitbit` stay `""` (the new XML default) — where "N% to goal" used to render

Net effect on those paths: the PR *removes* the percentage line without delivering the rings that were meant to replace it.

**Suggested fix:** move the metrics/ring render into `ChartManager.updateRing` (it already has the context and the view refs), so every caller is covered — or route those five sites through the MainActivity wrappers.

---

## Worth verifying before the on-device pass

### 3. `activities/tracker/calories` is total burn, including BMR

`NxFitbitHelper.getActivityByDate` builds `user/-/activities/tracker/<metric>/date/<date>/1d.json`. Fitbit's `calories` time series is **total** daily calories including BMR (~1800–2800/day), whereas the HC path uses `ActiveCaloriesBurnedRecord` — active only (~200–400).

Two consequences:
- Fitbit mode will show a number ~6–10× HC's for identical activity — inconsistent semantics between modes for the same on-screen label
- the calorie ring (`calVal / 500f`) will be pinned at 100% from mid-morning, every day, for every user

`activityCalories` is the activity-only resource if you want parity with HC.

### 4. `getFieldFromProfile` never returns null — it returns the literal string `"An error occurred"`

```java
String distUnit = fitbit.getFieldFromProfile("distanceUnit");
boolean miles = distUnit != null && distUnit.toUpperCase().contains("US");
```

The null guard can't fire (`NxFitbitHelper.java:184` initialises the return to that sentinel). On a profile-fetch failure the sentinel doesn't contain "US", so it silently assumes metric — a US account's miles get stored as kilometres, a **1.61× understatement**, presented as a real value with no `≈`.

**Suggested fix:** check for the sentinel (or match `METRIC` explicitly) and return `-1` on failure so it degrades to the step-derived estimate rather than a wrong "real" number.

Also `toUpperCase()` without a locale — use `Locale.ROOT`. The app ships `values-tr`, and locale-sensitive case mapping is precisely the class of bug that bites in Turkish.

---

## Minor

- `renderDashboardRings` now runs on **every sensor tick** via `displayActivityChart` (`MainActivity.java:1826`, `animate=false`): 3× `findViewById` + `getSharedPreferences` + `CompanionUtil.resolveCompanion` + `Calendar.getInstance` + `getColor` + `invalidate()` on a `LAYER_TYPE_SOFTWARE` view. Worth caching the view refs / hoisting the prefs read.
- Double `editor.apply()` back-to-back at :1160 and :1166 — the second batch can join the first.
- `sumFitbitTrackerMetric` returns `-1` for an empty array but `0` when the array holds zeros. That's the right call, but it means a genuine zero-distance day renders "0 km" with no `≈` while a fetch failure renders an estimate — worth a comment so the asymmetry is intentional on the record.
- Two extra Fitbit API calls per sync (1 → 3). Comfortable against the 150 req/hour/user limit, just noting the change.

mcfarhat:
### ⚠️ Heads-up on the local working tree (not about this PR's code)

While reviewing I ran `git checkout origin/feat/unified-dashboard-rings -- .` before noticing the branch was already checked out. That was unnecessary, and it restores **all** tracked files from the ref — discarding any uncommitted edits to them.

Earlier in the same session the working tree had uncommitted modifications to:

- `.idea/misc.xml` (1 line removed)
- `app/unofficial_strings.xml.example` (1 line changed)

If those edits were still present when I ran that command, they're gone. They were never staged or stashed, so git can't recover them.

**Please check whether you're missing anything in those two files.** `app/unofficial_strings.xml.example` is the one that matters — `.idea/misc.xml` is IDE state. VS Code's Timeline view (right-click the file → Timeline / "Local History") may still hold the prior contents if you need them back.

The tree is otherwise clean and identical to `4572d53`; no source files were affected.

mcfarhat:
Thanks — both blockers were real. Addressed in 7c38b58:

**Blocker 1 (Fitbit staleness).** Added a `fitbitMetricsDate` stamp on sync; `updateFitbitDashboardRings` now coerces a cache not stamped with today's date back to `-1`, so a stale distance/calorie shows as a `≈` estimate instead of yesterday's number presented as today's real data. (The fetch still only runs on the OAuth-callback path — same as `fitbitSyncCount` — so this is the freshness guard you suggested rather than moving the fetch; happy to relocate the fetch to the ordinary sync in a follow-up.)

**Blocker 2 (bypassed wrappers).** Moved the ring+metrics render into `ChartManager.displayActivityChart` / `displayActivityChartFitbit` (they already hold the `MainActivity` context), and removed the now-duplicate calls from MainActivity's wrappers. `TrackingManager.useDefaultTrackingMethod`'s five direct `ChartManager` calls now render the rings and the metrics line, so the startup/fallback path no longer drops the percentage without delivering the rings.

**#3 (calories incl. BMR).** Switched Fitbit calories from `tracker/calories` (total, ~1800–2800) to `activityCalories` (active only), for parity with HC's `ActiveCaloriesBurnedRecord` — otherwise the `calVal/500` ring pinned at 100% all day. Added `NxFitbitHelper.getActivityResource` since `activityCalories` isn't a `tracker/` resource.

**#4 (`distanceUnit` sentinel + locale).** Correct — `getFieldFromProfile` returns `"An error occurred"`, never null. Now I only convert on an explicit `METRIC`/`US` match and otherwise leave `-1` (→ estimate) rather than guessing metric; `toUpperCase(Locale.ROOT)` for Turkish-safe casing.

**Minors.** Collapsed the double `editor.apply()`; documented the `-1` (unavailable → estimate) vs `0` (genuine zero → real "0") asymmetry in `sumFitbitActivityResource`. The per-sensor-tick cost of `renderDashboardRings` I've left for now — the `AuraView` already self-invalidates each frame for the breathing animation, so the extra `setActivityRings` isn't the dominant cost — but noted for a caching pass if it shows up.

On the working-tree heads-up: thanks for the flag — those two files (`.idea/misc.xml`, `unofficial_strings.xml.example`) are gitignored-style noise I'd deliberately kept out of every commit this session, so nothing of substance was lost. I'll double-check the `.example` regardless.

mcfarhat:
## Re-review — `7c38b58`

All four points from the previous round are properly fixed. `compileDebugJavaWithJavac` passes (pre-existing Java 8 warnings only). I traced the two structural changes since they had the most room to go wrong:

- **Blocker 1 (staleness)** — `fitbitMetricsDate` is stamped in the same `apply()` batch as the values, and `updateFitbitDashboardRings` gates on it, so a stale cache degrades to `-1` → `≈` estimate. Reordering the render to *after* the pref writes also means the first paint on the OAuth return uses the fresh values rather than the previous cache. Correct.
- **Blocker 2 (coverage)** — moving the render into `ChartManager` is the right call and the details hold up: both new calls sit **inside** the existing `runOnUiThread` lambdas (so dropping the explicit `runOnUiThread` from the OAuth path is safe), the `instanceof MainActivity` guard is sound, the now-duplicate calls were removed from MainActivity's wrappers so nothing double-renders, and `renderDashboardRings` never calls back into `ChartManager`, so there's no recursion. `TrackingManager`'s five direct calls are covered.
- **#3** — `activityCalories` is the right resource for HC parity, and the response-key mapping `"activities-" + path.replace("/","-")` yields `activities-tracker-distance` and `activities-activityCalories`, both matching the time-series API. Keeping `tracker/distance` (device-recorded) is consistent with `tracker/steps`.
- **#4** — the explicit `METRIC`/`US` match with `-1` on anything unrecognized is the safe direction… but see below, because the premise underneath it turns out to be wrong.

---

## 🔴 New blocker: the Fitbit distance conversion is inverted

We never send an `Accept-Language` header. `NxFitbitHelper.makeApiRequest` builds a bare signed request:

```java
final OAuthRequest activityrequest = new OAuthRequest(Verb.GET, fitBitAPIUrl + endPointUrl);
serviceCall.signRequest(accessToken, activityrequest);
response = serviceCall.execute(activityrequest);
```

Per Fitbit's [Application Design guide](https://dev.fitbit.com/build/reference/web-api/developer-guide/application-design/), the response unit system is chosen **solely** by that header, and the default when it's absent is metric:

| `Accept-Language` | Unit system | Distance |
|---|---|---|
| `en_US` | United States | **miles** |
| `en_GB` | United Kingdom | kilometers |
| *(not provided / misspelled)* | **Metric** | **kilometers** |

> "API calls reveal data and log resource values in one of the unit systems based on the value of the Accept-Language header."
> "The Accept-Language header value is case sensitive. Improper case or misspelling will result in the metric unit system."

And the profile's `distanceUnit` is documented as the user's *display* preference, not a description of the API response:

> "The Units fields (`distanceUnit`, `glucoseUnit`, etc.) represent the default unit settings configured for the user. Numerical values are returned in the unit system specified in the Accept-Language header."

So `activities-tracker-distance` always comes back in **kilometers**, regardless of the account's preference — but the new code does this:

```java
String distUnit = fitbit.getFieldFromProfile("distanceUnit");
String u = (distUnit == null) ? "" : distUnit.toUpperCase(java.util.Locale.ROOT);
if (u.contains("US")) {
    fbDistM = fbDistRaw * 1609.344f;      // ← treats kilometres as miles
} else if (u.contains("METRIC")) {
    fbDistM = fbDistRaw * 1000f;
}
```

**Effect:** every Fitbit account whose profile `distanceUnit` is `en_US` gets its distance inflated by **1.609×**. A real 5 km day is stored as 8046 m and rendered as `8.05 km` (or `5.00 mi` for a US app user, who should be seeing `3.11 mi`). Accounts set to `METRIC` are correct by luck — the branch that fires for them is the one that matches reality.

**Fix:** delete the profile lookup and the branch; the value is always km:

```java
float fbDistM = (fbDistRaw >= 0) ? fbDistRaw * 1000f : -1f;
```

Optionally make it explicit rather than relying on the default, in `makeApiRequest`:

```java
activityrequest.addHeader("Accept-Language", "metric");   // case-sensitive per Fitbit docs
```

That one-liner also retires the `getFieldFromProfile` sentinel handling, the `Locale.ROOT` casing concern, and the `en_GB`-degrades-to-an-estimate coverage gap from the last round — all three only existed to support a conversion that shouldn't happen. Worth noting if you add the header globally: it governs weight/height units too, and we request the `weight` scope.

**On the app's own metric/US setting** — that layering is already right and should stay as is. `Utils.isMetricSystem` (the `activeSystem` pref) is only read at render time by `Utils.formatDistance`; it must *not* be sent to Fitbit, since a user can perfectly well have a metric Fitbit account and a US display preference in Actifit. Canonical metres in storage + app-setting formatting at the edge is the correct design. The bug is purely that the ingest side guesses the wire unit instead of pinning it.

---

## Also new: the OAuth-return path now makes 4 blocking network calls on the main thread

The `returnUrl != null` block runs inline in `onCreate` on the UI thread, and the only reason ScribeJava's synchronous `serviceCall.execute()` doesn't throw `NetworkOnMainThreadException` is that `NxFitbitHelper`'s constructor sets `StrictMode.ThreadPolicy.permitAll()` (`NxFitbitHelper.java:60-61`).

This PR adds `tracker/distance` + `activityCalories` to that block, so it's now profile → steps → distance → calories: four sequential round-trips blocking `onCreate` on the Fitbit-connect return. Pre-existing pattern, but the PR doubles the exposure — a visible freeze or ANR on a slow connection. The previous revision wrapped the render in `runOnUiThread`, implying background intent; the reorder made it fully synchronous inline. Moving just the two new fetches to a worker and rendering through the existing UI-thread path restores that.

---

## Minor

- **`SimpleDateFormat("yyyyMMdd")` without a `Locale`** on both the write (`:1156`) and read (`updateFitbitDashboardRings`) sides. Self-consistent, so it works — but it diverges from the HC convention (`ProfileActivity.hcMetricsFresh` uses `Locale.ENGLISH`) and inherits a latent bug. Pin both to `Locale.ENGLISH`.
  - Adjacent and pre-existing, probably its own ticket: `TrackingManager.java:296` writes `hcMetricsDate` with the default locale while `ProfileActivity.java:212` reads it with `Locale.ENGLISH`. On locales whose default numbering system isn't `latn` (several `ar-*`, `fa`, `th` — we ship `values-ar`) those never match, so the profile ring legend silently always shows estimates.
- **HC is now the odd one out.** `ChartManager` renders rings + metrics for sensors and Fitbit, but `displayActivityChartHealthConnect` still doesn't — HC waits on `TrackingManager`'s async read, so `tv_step_pct_hc` is blank between first paint and that callback. The same cached-then-refresh treatment would make all three consistent.
- `updateRing` animates `step_ring`/`step_ring_fitbit` progress immediately before `renderDashboardRings` sets it `GONE` — wasted animation on a hidden view, every tick.
- `updateFitbitDashboardRings` allocates a `SimpleDateFormat` and re-reads prefs on every chart refresh; easy to hoist if the tick cost ever matters.

mcfarhat:
## Implementation notes for the distance-unit fix

Concrete patch for the blocker above, so it can go straight in. Two files, both small.

### 1. `NxFitbitHelper.makeApiRequest` — pin the response unit system

Right now the request carries no `Accept-Language`, so we're relying on Fitbit's undeclared default. Make it explicit so a future default change can't silently move every distance by 1.6×.

```diff
     //function handles processing actual calls to grab Fitbit data for syncing purposes
     JSONObject makeApiRequest(String endPointUrl) throws InterruptedException, ExecutionException, IOException {
         OAuth20Service serviceCall = getService();
 
         final OAuthRequest activityrequest = new OAuthRequest(Verb.GET, fitBitAPIUrl + endPointUrl);
+        // Pin the response unit system. Fitbit selects units *solely* from this header — the account's
+        // own distanceUnit profile field is a display preference and does not affect API responses.
+        // "metric" => distance in km, weight in kg. Value is case-sensitive; a typo silently falls back
+        // to metric, which is why we normalise to metres at the call sites regardless.
+        // https://dev.fitbit.com/build/reference/web-api/developer-guide/application-design/
+        activityrequest.addHeader("Accept-Language", "metric");
         serviceCall.signRequest(accessToken, activityrequest);
```

Note this is global to every Fitbit call. That's intended — it also fixes the units for anything added later under the `weight` scope. Nothing currently reads weight/height, so there's no existing consumer to migrate.

### 2. `MainActivity` (~line 1143) — delete the guess, always convert km → m

```diff
-                            // also fetch Fitbit's own distance + activity calories for the multi-metric
-                            // rings. activityCalories (not tracker/calories, which includes BMR) keeps
-                            // parity with Health Connect's active-calories. Distance is returned in the
-                            // account's unit, so convert to metres via the profile's distanceUnit.
+                            // also fetch Fitbit's own distance + activity calories for the multi-metric
+                            // rings. activityCalories (not tracker/calories, which includes BMR) keeps
+                            // parity with Health Connect's active-calories. Distance always arrives in
+                            // km (Accept-Language: metric, set in NxFitbitHelper.makeApiRequest), so the
+                            // conversion to metres is unconditional.
                             float fbDistRaw = sumFitbitActivityResource(fitbit, "tracker/distance", targetDate);
                             float fbCal = sumFitbitActivityResource(fitbit, "activityCalories", targetDate);
-                            float fbDistM = -1f;
-                            if (fbDistRaw >= 0) {
-                                String distUnit = fitbit.getFieldFromProfile("distanceUnit"); // "METRIC"/"en_US" or a sentinel on failure
-                                String u = (distUnit == null) ? "" : distUnit.toUpperCase(java.util.Locale.ROOT);
-                                if (u.contains("US")) {
-                                    fbDistM = fbDistRaw * 1609.344f;      // imperial account -> miles
-                                } else if (u.contains("METRIC")) {
-                                    fbDistM = fbDistRaw * 1000f;          // metric account -> km
-                                }
-                                // else: unknown/failed profile lookup -> leave -1 so distance degrades
-                                // to a "≈" estimate rather than guessing the wrong unit
-                            }
+                            float fbDistM = (fbDistRaw >= 0) ? fbDistRaw * 1000f : -1f;
```

`-1` still propagates on a failed/empty fetch, so the `≈` estimate path is unchanged.

### What this retires

Three items from the previous round existed only to prop up the conversion branch, and all go away with it:

- the `getFieldFromProfile` `"An error occurred"` sentinel handling
- `toUpperCase(Locale.ROOT)` / Turkish-casing concern
- `en_GB` (and any other locale string) silently degrading to a `≈` estimate — the earlier "watch whether estimates are common" note is moot

No other consumer is affected: `PostSteemitActivity`'s Fitbit sync only reads `activities-tracker-steps`, which is unitless.

### Verification

The wrong and right values differ by exactly 1.609344×, so this is easy to check without a US-preference account:

1. On a Fitbit-connected account, note today's distance in the Fitbit app.
2. Reconnect Fitbit in Actifit (the fetch only runs on the OAuth-callback path) and compare the `📏` value on the dashboard.
3. Metric accounts should be unchanged by this patch; accounts whose profile `distanceUnit` is `en_US` should **drop** by ~38% (÷1.609) and now match the Fitbit app.
4. Worth logging `fbDistRaw` once during the check — if the API is honouring the new header it should be a plain kilometre figure matching the Fitbit app's km reading.

Toggling Actifit's own metric/US setting should change only the rendering (`km` ↔ `mi`), never the stored metres — that's the layering guard that this bug slipped past.

mcfarhat:
Great catch — the conversion was genuinely inverted, and the root cause (Accept-Language, not `distanceUnit`, governs the wire unit) is the important bit. Applied your patch in e35de7f:

**Distance unit.** Pinned `Accept-Language: metric` in `NxFitbitHelper.makeApiRequest`, and replaced the profile-based guess with the unconditional `fbDistRaw >= 0 ? fbDistRaw * 1000f : -1f`. So `activities-tracker-distance` is always km and always ×1000 → metres; `-1` still degrades to the `≈` estimate. As you noted, that retires the `getFieldFromProfile` sentinel handling, the `Locale.ROOT` casing concern, and the `en_GB`-degrades gap in one go — all three only existed to prop up a conversion that shouldn't happen. Canonical metres in storage + app-setting formatting at the edge (`Utils.formatDistance`) is untouched.

**Main-thread network.** Moved the two new fetches (`tracker/distance` + `activityCalories`) to a worker thread that caches the values and re-renders via `runOnUiThread`; steps still paint immediately and the metrics fill in when the fetch returns. That takes the OAuth-return `onCreate` back off the critical path for the additions (the pre-existing profile+steps calls are still synchronous — happy to lift those too in a separate pass, but left them out of this PR's scope).

**Locale.** Pinned `SimpleDateFormat("yyyyMMdd", Locale.ENGLISH)` on both the write and the freshness-read sides.

Verification path for when a Fitbit account is available: metric accounts unchanged, `en_US`-preference accounts drop ~38% (÷1.609) to match the Fitbit app — and toggling Actifit's own metric/US setting only flips `km`↔`mi` at render, never the stored metres.

**Deferred (acknowledged):**
- The pre-existing `TrackingManager:296` vs `ProfileActivity:212` `hcMetricsDate` locale mismatch (default-locale write vs `Locale.ENGLISH` read) is a real latent bug for non-`latn` locales but sits outside this PR — worth its own ticket.
- HC's cached-then-refresh consistency (it renders via `TrackingManager`'s async read, not `ChartManager`), the wasted `step_ring` animation before it's hidden, and hoisting the `SimpleDateFormat`/prefs read in `updateFitbitDashboardRings` — all minor; can fold into a follow-up if you'd like them in now.

PR #85, opened 2026-08-05
Signup wizard — REWORK REQUIRED (reopens reverted #81)

mcfarhat:
## Status: ⛔ DO NOT MERGE — sent back for rework

PR #81 was merged into `develop` but **reverted** (commit `6cd8cc3`) because the signup wizard cannot complete a real signup against the live backend (`actifit-bot` / `api.actifit.io/confirmPayment`). This PR returns the work to `develop` as an open MR, rebuilt on **current** develop so it can be reworked in isolation.

## What must be fixed (verified against the backend `app.js` / `utils.js`)

1. **Auth token field name (blocks everything).** The client sends the token under the generic `sec_param` key, but `POST /confirmPayment` gates on a field literally named **`confirm_payment_token`** (`app.js:8496`). With the wrong key the server returns `{}` and the app polls until `payment_timeout` — this blocks **both** the promo and paid paths.
2. **Missing `steem_invest` + no USD→HIVE conversion.** The on-chain check does `sentAmount >= parseFloat(req.query.steem_invest) - 0.1` (`utils.js:757`). The app sends only `usd_invest` and never converts to HIVE, so the comparison is against `NaN` and never matches. Convert the USD cost to a live HIVE amount (backend already exposes a coingecko hive-price helper, `app.js:691`) and send it as `steem_invest`.
3. **Currency hardcoded to `HIVE`.** The UI offers "HIVE or HBD" but the client always sends `sent_cur: "HIVE"`, while the server requires `sentCur === req.query.sent_cur` (`utils.js:758`). HBD payments can never confirm. Send the currency the user actually paid with.
4. **Missing `afit_reward`.** `storeSignupTransaction` stores `parseFloat(req.query.afit_reward)` (`app.js:6981`); the app never sends it, so the signup reward is recorded as `NaN`.

## Client-side polish (from code review)
5. The payment **memo regenerates on every entry to step 3** (`SignupActivity.preparePaymentInfo`), so a Back/forward navigation orphans an already-sent payment. Guard it (`if (generatedMemo.isEmpty())`).
6. The memo uses standard Base64 (can contain `+` / `/`) but users must **type it by hand** into a wallet — use a hand-entry-friendly alphabet.

Detailed feedback also provided to the developer on Trello.

Reverted-by: `6cd8cc3` · Original PR: #81


mcfarhat:
@Adam-Diab1 the rework genuinely fixed **5 of the 6** stated requirements, and the manifest regression from #81 is gone — I diffed `uses-permission` against `develop` and they're identical. Credit where due:

- `steem_invest` with live USD→crypto conversion — matches the production web client exactly (same `toFixed(3)` string form).
- `sent_cur` now follows the HIVE/HBD toggle, so `utils.js:758`'s comparison will match for HBD.
- `afit_reward` is a **numerically faithful** port of the web's `getMatchingAFIT()` — I checked the lots/cap/floor maths and the AFIT price source against the live API.
- Memo regeneration guarded, and the hand-entry alphabet (no I/O/0/1) is a nice touch.
- Master-password handling is genuinely good: `SecureRandom` 40 bytes → ~300 bits, **never logged** anywhere, HTTPS-only, never persisted. No secrets committed, no `.idea/`, resources clean (zero duplicate keys across all four string files).

**But the headline question — can it complete a real signup against the live backend — is still no.** Three independent blockers, and I want to be direct: a successful debug build can't tell you this, so the flow can't have been exercised end-to-end.

**🔴 1. `confirm_payment_token` carries the wrong secret.** `SignupActivity.java:620` sends `R.string.sec_param_val`. But `sec_param_val` is `config.verifyPostToken`, which gates a *different* endpoint (`GET /appendVerifiedPost`, `app.js:8794`). The payment endpoint checks `config.confirmPaymentToken` (`app.js:8496`) — a different config key. Mismatch → `res.send('{}')` → no `accountCreated` → the app polls forever. **That is exactly the #81 failure.** There's no `confirm_payment_token` entry anywhere in the repo and this PR doesn't add one to `unofficial_strings.xml.example`, so no build can currently supply the right value.

⚠️ Before you just add a new string: the production **web client never sends this token from the browser** — it POSTs to `/api/proxy/confirmPayment` and a server-side proxy injects it. Shipping a bearer secret inside an APK is strictly worse, and it's compounded because the server mints `afit_reward` from the **client-supplied** value (`app.js:6981`, `:7054`, `:7093`). I'm raising the proxy-vs-token decision with Dr. Mohammad — **hold on this one until we come back to you.**

**🔴 2. The poll/timeout machinery is dead code.** `utils.js:730` only ever *resolves* — it polls the chain every 5s and never rejects — while `app.js:8504` writes a keep-alive byte every 6s. Volley's `setReadTimeout(30000)` is a *per-read* timeout, so the socket never trips. Neither listener fires, `pollHandler.postDelayed` never runs, and `MAX_POLL_ATTEMPTS` / `payment_timeout` (`:57`, `:592-596`) are unreachable. The user gets an unbounded "Verifying payment…" spinner. You need a real client-side deadline (`Handler` + `request.cancel()`).

**🔴 3. Paid-but-account-creation-failed is swallowed.** `app.js:8593` returns `{paymentReceivedTx: "<txid>", accountCreated: false}` when the payment **was** found but creation failed. You read only `accountCreated` (`:628-642`), so this is treated as "not paid yet" and loops forever — **the user has spent real HIVE and sees a spinner.** Check `paymentReceivedTx` and surface a terminal error with support guidance.

**Also must-fix:**
4. **Omit `referrer` when blank** (`:613`). You send `""`; the web omits the key. Backend guards (`app.js:6962`, `:6995`) all pass on `""`, so every referrer-less signup writes a `token_transactions` row with `user: ""` and fires a notification at an empty username.
5. **Username regex isn't Hive's** (`:254`, `:377`). `^[a-z][a-z0-9\-.]{2,15}$` accepts `ab-`, `a..b`, `abc--d`. `get_accounts` returns empty for those (reads as "available"), the user pays, and `account_create` fails on chain → straight into blocker 3. Hive needs per-dot-segment rules. Your helper text also says "lowercase, numbers, and hyphens" while the regex allows dots.
6. **Rotation or back destroys the master password after payment.** No `configChanges`, no `onSaveInstanceState`, no `onBackPressed`. Rotate at step 3–4 and `generatedMasterPassword` resets — if the account was already created, the user's only credential is gone permanently. Lock orientation and/or persist state.
7. **Guard async callbacks** with `isFinishing()`/`isDestroyed()` (`:627-643`, `:396-419`) and **cancel the Volley queue in `onDestroy`** (`:741`) — with the server holding the connection open, those lambdas pin the Activity for the process lifetime.
8. **Render payment instructions when the price fetch fails** (`:552`) — currently the user gets a Toast and an empty panel: no recipient, no memo, no amount.
9. **Drop the hardcoded fallback rates** (`:59-60`) — a fabricated `steem_invest` of `20.000` gets persisted for promo signups where nothing was paid. Send `0` on the promo path.
10. **Extract hardcoded strings** (`activity_signup.xml:994, 1037, 1127`; `SignupActivity.java:552-559`) — this regresses the i18n work from #86, and the wizard is currently English-only. Also guard `setupTermsLinks` (`:196-200`): `indexOf` returns `-1` on any translation, and `setSpan(-1, …)` throws — the first translator ships a crash.
11. **Compose URLs from `@string/live_server`** (`:62-68`) rather than hardcoding `https://api.actifit.io/…`, so `test_server`/`test_mode` works.

**Nits:** add `FLAG_SECURE` and `inputType="textPassword"` on the master-password screen (it currently renders unmasked and appears in screenshots/recents), and set `EXTRA_IS_SENSITIVE` on the clipboard copy so Android 13+ doesn't preview it; the memo prefix is `signup:` while production memos are `acmm…`; the promo-failure message says "invalid promo code" even when the promo was valid and already consumed; `catch (JSONException)` at `:622` doesn't `return`.

Real progress here — the contract work is close and the maths is right. Wait on the token decision before the next push.

mcfarhat:
@Adam-Diab1 decision on the token question from my last comment — **we're going with the server-side proxy (option a)**, mirroring how the production web client works. So please **don't** add a `confirm_payment_token` string to the app.

**Why:** a token shipped in an APK is extractable by anyone who downloads it, and the backend mints `afit_reward` from the **client-supplied** value (`app.js:6981`, `:7054`, `:7093`) and trusts `promo_code`. So a leaked token means arbitrary AFIT minting and burned promo codes. The web avoids this precisely by never exposing it to the client — it POSTs to `/api/proxy/confirmPayment` and the proxy injects the secret server-side, behind reCAPTCHA.

**What that means for this PR:**

- The app should call an **app-facing proxy endpoint** instead of `api.actifit.io/confirmPayment` directly, and send **no** payment token at all.
- That proxy needs to exist on our side — it doesn't yet for the app. **This part is on us, not you**, and Dr. Mohammad and I will sort the endpoint and how the app authenticates to it. I'll come back to you on this card with the endpoint and the contract.
- Until then, **leave the token line as-is** — don't work around it. Everything downstream of that call (the body it sends) is already correct, so this should end up being a URL change plus dropping one field.

**In the meantime, all the other items are unblocked and worth doing now** — none of them depend on the proxy decision:

2. Real client-side deadline (`Handler` + `request.cancel()`) — the backend keep-alives forever, so Volley's read timeout never trips and your whole poll/timeout path is unreachable today.
3. Handle `paymentReceivedTx` set with `accountCreated:false` as a terminal error — right now a user who has genuinely paid loops forever.
4. Omit `referrer` (and `email`) when blank.
5. Real Hive username rules at `:254`/`:377`.
6. Survive rotation / back / process death after step 3 — this is the one I'd do first. Losing the master password after the account exists is unrecoverable for that user.
7. Guard async callbacks with `isFinishing()`/`isDestroyed()` and cancel the Volley queue in `onDestroy`.
8. Render payment instructions even when the price fetch fails.
9. Drop the hardcoded fallback rates; send `0` on the promo path.
10. Extract the hardcoded strings and guard `setupTermsLinks` against `indexOf == -1`.
11. Compose URLs from `@string/live_server` — that'll matter for the proxy work too, so worth doing now.

Plus the master-password screen nits: `FLAG_SECURE`, `inputType="textPassword"`, and `EXTRA_IS_SENSITIVE` on the clipboard copy.

Ping me when those are in and I'll re-review; we'll wire the proxy in as a follow-up once it's ready.

mcfarhat:
## Review at `0ead7745` — changes requested

Reviewed the full diff, built it, and verified the mainnet run independently. Summary first: **the hard parts are done well and the remaining work is small and mostly one-line.** Six items to fix before merge.

I built this rather than reading it: supplied the gitignored files from the in-repo `.example` templates, then `./gradlew compileDebugJavaWithJavac` and `./gradlew assembleDebug` both **BUILD SUCCESSFUL**, producing a real debug APK.

**This is a genuine rework, not #81 re-pushed.** `git merge-base` against develop is the current tip, `SignupActivity` went 483 → 1132 lines, and `SignupState`/`SignupStateStore` don't exist in #81 at all. Rework items 2–6 from the PR body are all genuinely addressed.

### The master password design — decision recorded, not a blocker

`SignupActivity.java:810` sends `new_pass` to the server, and you flagged that yourself in the comment rather than hiding it. Mohammad has reviewed this and **accepted it for now**, on the basis that transport is HTTPS and it matches how the existing web signup already works. Flagging it openly was the right call.

It stays a known trade-off rather than a settled design: the master password derives owner, active, posting and memo keys, so the ideal shape is the client deriving keypairs locally and sending only public keys. That's a backend change, so it's a follow-up card, not your PR.

### Fix before merge

**1. `FLAG_SECURE` is missing on the screen that displays the master password.** Not set anywhere in `app/src/main`. Right now that screen can be screenshotted, screen-recorded, and appears in the recent-apps thumbnail. One line in `onCreate`.

**2. The clipboard copy isn't marked sensitive.** `SignupActivity.java:1042-1048` — no `ClipDescription.EXTRA_IS_SENSITIVE`. With minSdk 28 / targetSdk 35, Android 13+ renders the clipboard preview on screen, so the master password is shown in cleartext at the moment the user copies it. One line.

**3. The password field is never actually masked.** `activity_signup.xml:279-291` sets `app:endIconMode="password_toggle"` on the `TextInputLayout`, but the `TextInputEditText` has no `android:inputType="textPassword"`. Without a password inputType the toggle does nothing and the value renders in plaintext at all times. (`android:editable="false"` on :289 is also deprecated and ignored — use `android:focusable="false"` or set it in code.)

**4. A leftover recovery file permanently breaks auto-login for the whole app.** `LoginActivity.java:184-190` gates `queryAPI` on `signupStateStore.exists()`. Past step 2, only "Go to Login" (`:1074`) and `cancelBeforeIrreversibleBoundary` (`:373`, reachable only at steps 1–2) delete the file — "Leave for now" (`:363-366`) deliberately does not.

Scenario: an existing logged-in user taps "Create account" out of curiosity, reaches step 3, backs out. From then on **every** app launch skips auto-login and drops them on the login screen with a "Resume signup?" prompt, forever. They also can't start a fresh signup, because `SignupActivity.onCreate` always restores the stale state and jumps back to step 3. Only escape is clearing app data.

Needs a discard path (see 7) and, separately, the recovery check must not gate `queryAPI`.

**5. The currency toggle isn't locked after payment.** `:150-165`. User sends 6.4 HIVE, then taps "HBD" while waiting. `selectedCurrency` flips and is persisted at `:163`, so every subsequent confirm sends `sent_cur: "HBD"` — which can never match their HIVE transfer. Account never created, money gone, no diagnostic. Disable the toggle once `recoveryState.requestSubmitted` is set, or as soon as the memo and amount are displayed.

**6. A user who has already paid gets stuck if the AFIT price endpoint is down.** `:799-806` refuses to submit unless `afitReward >= 0`, and the recovery branch at `:803` re-fetches the currency price but never calls `fetchAfitReward()`. Within one Activity session, "Check Payment" then fails with "Signup pricing is still loading" no matter how many times they try — after their money is gone. Leaving and re-entering recovers it via `:575-577`, but nothing tells them that. Add `fetchAfitReward()` to the retry branch.

### Worth doing, not blocking

**7.** No "discard this signup and start over" affordance anywhere. Pairs with 4.

**8.** An invalid promo code is indistinguishable from an unpaid signup. `error_invalid_promo` is defined in `signup_strings.xml:11` and never used — a rejected code falls through to "Your signup status is not known yet", so the user re-taps forever with no idea the code was refused.

**9.** The 60s deadline is mostly dead. `:890-893` sets a Volley timeout of 30 000 ms with 0 retries while `CONFIRM_PAYMENT_DEADLINE_MS` is 60 000 (`:56`), so Volley errors first and the string "reached its 60-second limit" (`signup_strings.xml:66`) is essentially never true. Pick one number.

**10.** The account-creation-failure dialog (`:1018-1030`) correctly says "do not send another payment", but shows no memo, no username and no contact route. The memo is the only thing linking their on-chain payment to the attempt — surface it there so support can act.

**11.** Rotation at steps 1–2 resets the wizard. There's no `onSaveInstanceState` override (only `onRestoreInstanceState` at `:616`), and nothing persists until the step-2→3 transition at `:316`. Rotate on step 2 and you're back at step 1 re-running the availability check. Steps 3–4 are fine because the encrypted store covers them.

**12.** Master password alphabet is standard Base64 (`:1037-1038`), so it can contain `+` and `/`. You fixed exactly this hazard for the memo at `:628` — the same reasoning applies more strongly to the credential the user has to transcribe by hand and later type into a login screen.

**13.** Username isn't auto-lowercased (`activity_signup.xml:83` is `inputType="text"`), so typing `AdamTest` fails validation with no hint that case is the problem — even though the value gets lowercased later anyway at `:590`. And the `startsWith("uid")` / 10-consecutive-digit rules at `:503` aren't Hive rules, aren't commented, and aren't mentioned in `username_requirement`.

**14.** Dead code: `handleNetworkError()` (`:1050-1066`) is never called, the `VolleyError` import (`:35`) is unused, `:656`'s `lots` computation is always 1 given the constants, and 14 strings in `signup_strings.xml` are leftovers from #81's removed email-verification flow.

**15.** Convention: the five endpoint URLs, the `app/confirmPayment` path, `actifit.signup` and `SIGNUP_COST_USD` are hardcoded Java constants, where this repo's convention is that endpoints live in `unofficial_strings.xml` (see `HiveRequests.java:96`). Not a leak — just means changing the signup price needs an app release.

### Verified clean

- **`greateck_team`: zero hits** — checked the diff, the full tree, and the entire git history. Retired as asked.
- **No hardcoded promo codes, tokens, API keys or credentials** anywhere in the diff. Promo and referrer are user-entered only.
- **No key leakage on anything you control**: no `Log.*`, no `printStackTrace`, no `System.out` in the three new files; no `putExtra`, so the password never touches an Intent; never written to `SharedPreferences`; `setSaveEnabled(false)` keeps it out of the instance-state Bundle; wiped on success.
- **Username validation** (`:499-515`) faithfully mirrors Hive's `is_valid_account_name` — 3–16 chars, `.`-split segments, each ≥3 starting `[a-z]`, no `--` — and runs before any network call. No wasted round-trips.
- **Your mainnet run checks out.** I pulled the transaction independently: one `create_claimed_account`, creator `actifit`, block 108985696, `account_auths: []` and `weight_threshold: 1` on owner, active and posting, three distinct keys plus a distinct memo key, `recovery_account: actifit`. Correct shape.

Note the wizard doesn't construct authorities at all — the backend derives the keys and builds the op — so that guarantee lives server-side and this PR can't enforce it. Nothing for you to change; recording it so nobody later assumes the client is the control point.

### What's genuinely good here

`SignupStateStore` is the strongest code in this PR and better than a lot of what surrounds it: AndroidKeystore AES/GCM with `setRandomizedEncryptionRequired(true)`, a versioned envelope, length-bounded reads so a corrupt file can't drive a huge allocation, `AtomicFile` with proper `failWrite`, `getNoBackupFilesDir()` chosen specifically because the manifest sets `allowBackup="true"`, plaintext zeroed after use, alias deleted on clear, and fail-closed with a dedicated checked exception. `SignupState.fromJson` validates the whole phase/flag state machine and rejects impossible combinations.

The idempotency discipline around the irreversible boundary is the hard part of this feature and you got it right: state persists *before* `queue.add` (`:839`), there's a request-generation counter plus single-flight guards, `onDestroy` cancels in flight, and on recreate the `REQUEST_SUBMITTED` phase reconciles against Hive rather than re-POSTing. Rotation mid-confirm cannot double-charge. You also modelled the exhausted-token case explicitly with a "payment received, do not send another payment" state — anticipating a failure most people discover in production.

Fix 1–6 and this is ready.

Adam-Diab1:
Pushed the six requested blockers in commit 5289278 (fix: address final signup review blockers).
This includes FLAG_SECURE, sensitive clipboard handling, password masking, recovery no longer blocking auto-login, guarded discard behavior, post-submission currency locking, and AFIT-price retry.
Build and git diff --check both pass. Ready for re-review.

________________________________

mcfarhat:
## Re-reviewed at `52892780` — one line stands between this and merge

Built it: `./gradlew assembleDebug` → **BUILD SUCCESSFUL in 1m 27s**, real 50 MB APK.

### Key handling — no regressions, checked first

This was my main worry on a late commit under time pressure, so I checked it before anything else:

- **Zero `Log.*` / `printStackTrace` / `System.out`** in the three files — and I grepped every *added* line in the whole PR diff, not just the current state. None introduced while fixing the six items.
- No `putExtra`, never written to `SharedPreferences`, `setSaveEnabled(false)` still at `:149`, wiped on success at `:1117-1119`.
- **`SignupStateStore.java` is byte-identical** to the version I approved — Keystore AES/GCM, `setRandomizedEncryptionRequired`, length-bounded reads, `AtomicFile` with `failWrite`, `getNoBackupFilesDir()`, plaintext zeroed, alias deleted on clear.
- **Idempotency re-derived rather than assumed.** State persists at `:872` before `queue.add` at `:933`; generation counter at `:877`; `onDestroy` cancels by tag; on recreate `onCreate:130-132` routes `REQUEST_SUBMITTED && !accountCreated` to a read-only reconcile. I traced every `persistRecoveryState` call site — `requestSubmitted` can never be reset to false, because the only `false` write is the step 2→3 transition and `applyRecoveryState` always lands on step 3 or 4 with no path back. **Rotation mid-confirm still cannot double-charge.**
- **`greateck_team`: zero hits** across the diff, the full tree, and all git history.

In a commit that was pure behavioural fixes at the end of the programme, adding no debug logging and leaving the crypto store untouched is the discipline that actually matters here.

### The six items

**1. `FLAG_SECURE` — fixed.** `:113`, after `super.onCreate` and **before `setContentView`**, so it covers the first frame and the recents thumbnail. Never cleared anywhere.

**2. Clipboard — fixed.** `:1079-1086`, correct API (`getDescription().setExtras()` with `EXTRA_IS_SENSITIVE`), null-safe on existing extras, correctly guarded on `TIRAMISU`.

**3. Password masking — fixed.** `activity_signup.xml:290-294` now has `inputType="textPassword"`, and you replaced the deprecated `android:editable` with `clickable`/`cursorVisible`/`focusable` — which I'd only mentioned in parentheses.

**4. Auto-login gate — the important half is fixed.** `LoginActivity.java:185-188` now calls `queryAPI` unconditionally, so a leftover recovery file no longer breaks auto-login for the whole app. No NPE from dropping `initializeItems()` there — it's still reached via `assessLogin` on both paths that show the login UI. But the discard path you added alongside it is unreachable — see below.

**5. Currency lock — partially fixed.** `updateCurrencyToggleState()` at `:763-768` disables the group *and* both children, which correctly handles the Material quirk where `setEnabled` doesn't propagate. That closes the in-flight case. It doesn't close the window that actually loses money — see C.

**6. AFIT price retry — fixed, and better than asked.** `:833` adds the `afitReward < 0` refetch, and you tightened the price refetch to fire only when it's genuinely missing. It now also covers the **promo** path, which previously refetched nothing and was permanently stuck — I hadn't called that out.

---

## Blocking — one hardcoded `true`

`SignupState.canSafelyDiscard()` requires `!irreversible`. But `SignupActivity.java:612` passes the **literal `true`** for `irreversible` on every write, and `:601` is the only `new SignupState(...)` in the codebase. So `canSafelyDiscard()` can never return true.

Repro: fresh install → Create account → step 2 → Proceed → back out → relaunch → "Resume signup?" → tap **Discard signup** → "Signup cannot be safely discarded", every time. Combined with `restoreRecoverableSignup()` always restoring, the user still can't start a fresh signup under a different username, and clearing app data remains the only escape.

You clearly intended otherwise — in the same commit you relaxed `SignupState.fromJson` to accept `READY_FOR_PAYMENT` with `irreversible=false`, a state you then never write. So the validator learned about a state the writer never produces.

Fix is one line at `:612`, e.g. `irreversible = !PHASE_READY_FOR_PAYMENT.equals(phase)`. One design thought first: at `READY_FOR_PAYMENT` the user may already have transferred on-chain without confirming, so discarding there should carry a strong warning rather than being silent.

---

## Two follow-ups — not blocking, and one is partly my fault

**B. The recovery dialog is now unreachable for exactly the user it was built for.** `LoginActivity:185-188` fires the async `queryAPI` and then shows the dialog synchronously in `onCreate`. On successful auto-login the callback runs `proceedMain()` → `startActivity` + `finish()` while the dialog is attached, so it's torn down with a `WindowLeaked` in logcat. An existing logged-in user with an abandoned signup sees a flash and lands in MainActivity, never able to tap Resume or Discard. Not a crash. Fix: show it from the `assessLogin` no-token path, or defer to MainActivity.

**C. Item 5's real money window is still open, and my spec is half the reason.** The toggle only locks once `requestSubmitted` is true. So: step 3, memo and amount displayed → user sends 6.400 HIVE from their wallet → returns → taps **HBD** (still enabled; `selectedCurrency` flips, amount recomputes, memo unchanged because `preparePaymentInfo` generates it once) → taps **Check Payment** → body carries `sent_cur: "HBD"` against a HIVE transfer. Server can't match it, `requestSubmitted` *then* flips true and locks the toggle on HBD permanently, and `handleNextStep` only reconciles against Hive and never re-POSTs. No in-app path back. Money stranded.

I offered "once `requestSubmitted` is set, **or** as soon as the memo and amount are displayed" and you implemented the first. The second isn't a clean hard-lock either, since the user legitimately picks currency at step 3. The right shape is probably a confirm-on-flip once the memo is on screen ("only switch if you haven't sent yet"), or moving the currency choice ahead of memo generation. That's a design change, not a one-liner, so it's a follow-up card rather than something to hold your finish on.

---

## Still open from the earlier non-blocking list

None of these moved, all still fine to leave: `error_invalid_promo` defined and never referenced; the `30000` retry policy at `:924` still contradicting `CONFIRM_PAYMENT_DEADLINE_MS = 60_000`; the failure dialog still showing no memo or contact route; no `onSaveInstanceState`; `generateKeys:1070` still standard Base64 with `+`/`/`; username not auto-lowercased; `handleNetworkError` still dead with an unused `VolleyError` import; the four price/terms URLs still hardcoded rather than in `unofficial_strings.xml`.

## Not verified

No emulator, so `FLAG_SECURE` actually blanking screenshots, the password toggle rendering, clipboard-preview suppression, and B's `WindowLeaked` sequence are derived from the code and platform contracts, not observed. No end-to-end signup was run, so the idempotency argument is static re-derivation.

---

Fix the `true` at `:612` and I'll approve. The three security items are done precisely — right API, right SDK guard, right lifecycle placement — and item 6 caught a path I'd missed.

Adam-Diab1:
Fixed the blocking irreversible issue. READY_FOR_PAYMENT is now discardable, while REQUEST_SUBMITTED, ACCOUNT_CREATED, and ACCOUNT_CREATION_FAILED remain irreversible. Build and git diff --check pass. Pushed the fix in <commit>. I left B/C and the other non-blocking follow-ups untouched.

________________________________

mcfarhat:
🤖 REVIEW — wizard verified working end-to-end; ready for merge review.

**Root cause of "stuck at verifying payment" was NOT this client** — it was a server-side concurrency bug in actifit-bot (`confirmPaymentReceived` shared a module-level `setInterval` handle → timed-out polls became immortal zombies that killed every new signup's poll). Fixed + deployed in actifit/actifit-bot#49 (merged to master, live on both servers + Heroku). With that fixed, a real signup completed cleanly (`check funds → found → account created:true`) and the account was created on-chain.

**Final commit (`053f5ed`) on top of the reviewed wizard:**
- `FLAG_SECURE` now applied **per-step** — only step 4 (keys screen: master password + posting key) is screenshot-blocked; the rest of the wizard is capturable again (fixes the app-wide "screenshots not allowed" regression).
- Close the soft keyboard on **"Check Availability"** so the result under the field is visible.
- Removed the temporary `SignupDebug` diagnostic logging (it was only there to trace the server stall).

Device-tested on a physical device through full signup → payment → account creation → keys screen. No client-side blockers remain.

PR #86, opened 2026-08-06
Fix hardcoded step-goal strings (i18n) in ChartManager

AsmaaZeid:
## What this does
Fixes a genuine i18n bug in `ChartManager.java`'s `updateRing()` — two user-facing strings were built directly in Java with hardcoded English text instead of string resources:
- `tvGoal.setText("/ " + ... + " steps")`
- `tvPct.setText(progress + "% to goal")`

This affected every step-progress ring in the app (both the regular activity chart and the Health Connect chart use this same method), in every language — users always saw literal English regardless of locale.

## Changes
- Added `step_goal_format` and `step_pct_to_goal_format` string resources with placeholders
- Updated `ChartManager.java` to use `context.getString(...)` instead of hardcoded concatenation
- Translated both strings across all 13 locales

## Notes
- Fixed an unescaped apostrophe in the Italian translation (`all'obiettivo` → `all\'obiettivo`)
- Flagging `yo` (Yoruba) as not native-speaker verified

### DIFF
--- a/app/src/main/java/io/actifit/fitnesstracker/actifitfitnesstracker/ChartManager.java
+++ b/app/src/main/java/io/actifit/fitnesstracker/actifitfitnesstracker/ChartManager.java
@@ -136,8 +136,8 @@ public class ChartManager {
             tvCount.setText(java.text.NumberFormat.getInstance().format(steps));
             tvCount.setTextColor(color);
         }
-        if (tvGoal != null) tvGoal.setText("/ " + java.text.NumberFormat.getInstance().format(DAILY_GOAL) + " steps");
-        if (tvPct != null) tvPct.setText(progress + "% to goal");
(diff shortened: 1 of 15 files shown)


mcfarhat (approved):
Approving. Verified against the code, not just the diff:

- Format-arg types match the specifiers: `step_goal_format` `%1$s` ← `NumberFormat.format(DAILY_GOAL)` (String); `step_pct_to_goal_format` `%1$d` ← `int progress` (ChartManager.java:123). No IllegalFormatConversionException risk.
- `context` is a valid field and available in `updateRing`; the existing `tvGoal != null` / `tvPct != null` guards are preserved (the HC path passes null tvPct).
- Placeholder consistency confirmed across all 14 files — every locale uses exactly `%1$s` (goal) and `%1$d%%` (pct); `%%` escaping and the Italian apostrophe escape are correct.

Real bug (both the activity and Health Connect charts route through `updateRing`, so all locales showed English). Good to ride along in v0.14.2. Non-blocking: pt/pt-rBR are identical; `yo` is machine-translated (already flagged).

PR #87, opened 2026-08-07
AI Assistant enhancements: persist session, fix Insert button state, rename Accept to Insert

AsmaaZeid:
## What this does
Addresses Dr. Farhat's feedback after testing the AI Assistant popup:

1. **Session persistence** — Closing the AI popup previously destroyed the entire chat history (`chatHistory`/`chatAdapter` were local variables recreated on every open). Moved both to class-level fields on `PostSteemitActivity`, lazily initialized only once, so the conversation now persists across close/reopen cycles for the lifetime of the activity.

2. **Insert button state on reopen** — Previously the "Insert" button was unconditionally disabled every time the popup opened, even if a valid AI response already existed in history from a prior session. Now checks the (persisted) history on open and enables the button correctly if there's a usable AI response.

3. **Renamed "Accept" → "Insert"** — Updated the button label across all 13 locales, per feedback.

## Notes
- Scope limited to `PostSteemitActivity.java` (persistence logic) + all locale `strings.xml` (label rename)

### DIFF
--- a/app/src/main/java/io/actifit/fitnesstracker/actifitfitnesstracker/AiService.java
+++ b/app/src/main/java/io/actifit/fitnesstracker/actifitfitnesstracker/AiService.java
@@ -448,15 +448,18 @@ public class AiService {
 
     public CancellableRequest generateChatResponse(List<ChatMessage> history, final TextResponseCallback callback) {
         final Call[] currentCall = new Call[1];
-        generateChatResponseWithRetry(history, callback, 0, currentCall);
+        final boolean[] canceled = {false};
+        generateChatResponseWithRetry(history, callback, 0, currentCall, canceled);
         return () -> {
+            canceled[0] = true;
             if (currentCall[0] != null) {
                 currentCall[0].cancel();
             }
         };
     }
 
-    private void generateChatResponseWithRetry(List<ChatMessage> history, final TextResponseCallback callback, final int attempt, final Call[] currentCall) {
+    private void generateChatResponseWithRetry(List<ChatMessage> history, final TextResponseCallback callback, final int attempt, final Call[] currentCall, final boolean[] canceled) {
+        if (canceled[0]) return;
         List<Map<String, Object>> contents = new ArrayList<>();
         for (ChatMessage message : history) {
             Map<String, Object> turn = new HashMap<>();
@@ -464,7 +467,6 @@ public class AiService {
             turn.put("parts", List.of(Map.of("text", message.getText())));
             contents.add(turn);
         }
-
         Map<String, Object> requestBodyMap = new HashMap<>();
         requestBodyMap.put("contents", contents);
         String requestJson = gson.toJson(requestBodyMap);
@@ -474,7 +476,6 @@ public class AiService {
                 .addHeader("x-goog-api-key", API_KEY)
                 .post(requestBody)
                 .build();
-
         Call call = client.newCall(request);
         currentCall[0] = call;
         call.enqueue(new Callback() {
@@ -496,7 +497,10 @@ public class AiService {
                 } else if (response.code() == 503 && attempt < 3) {
                     long delayMs = (attempt + 1) * 1500L;
                     new android.os.Handler(android.os.Looper.getMainLooper()).postDelayed(
-                            () -> generateChatResponseWithRetry(history, callback, attempt + 1, currentCall),
+                            () -> {
+                                if (canceled[0]) return;
+                                generateChatResponseWithRetry(history, callback, attempt + 1, currentCall, canceled);
+                            },
                             delayMs
                     );
                 } else {
--- a/app/src/main/java/io/actifit/fitnesstracker/actifitfitnesstracker/AiService.java
+++ b/app/src/main/java/io/actifit/fitnesstracker/actifitfitnesstracker/AiService.java
@@ -288,6 +288,10 @@ public class AiService {
         void onFailure(String errorMessage);
     }
 
+    public interface CancellableRequest {
+        void cancel();
+    }
+
     public void generateDashboardInsight(int todaySteps, int streakDays, int avgSteps7d, final TextResponseCallback callback) {
         String prompt = "You are a fitness coach. Give one motivating, personalized fitness insight in exactly 1-2 short sentences. "
                 + "The user has walked " + todaySteps + " steps today, has a " + streakDays
@@ -442,11 +446,17 @@ public class AiService {
         });
     }
 
-    public void generateChatResponse(List<ChatMessage> history, final TextResponseCallback callback) {
-        generateChatResponseWithRetry(history, callback, 0);
+    public CancellableRequest generateChatResponse(List<ChatMessage> history, final TextResponseCallback callback) {
+        final Call[] currentCall = new Call[1];
+        generateChatResponseWithRetry(history, callback, 0, currentCall);
+        return () -> {
+            if (currentCall[0] != null) {
+                currentCall[0].cancel();
+            }
+        };
     }
 
-    private void generateChatResponseWithRetry(List<ChatMessage> history, final TextResponseCallback callback, final int attempt) {
+    private void generateChatResponseWithRetry(List<ChatMessage> history, final TextResponseCallback callback, final int attempt, final Call[] currentCall) {
         List<Map<String, Object>> contents = new ArrayList<>();
         for (ChatMessage message : history) {
             Map<String, Object> turn = new HashMap<>();
@@ -465,11 +475,15 @@ public class AiService {
                 .post(requestBody)
                 .build();
 
-        client.newCall(request).enqueue(new Callback() {
+        Call call = client.newCall(request);
+        currentCall[0] = call;
+        call.enqueue(new Callback() {
             @Override public void onFailure(Call call, IOException e) {
+                if (call.isCanceled()) return;
                 callback.onFailure("Network error: " + e.getMessage());
             }
             @Override public void onResponse(Call call, Response response) throws IOException {
+                if (call.isCanceled()) return;
                 String responseBody = response.body().string();
                 if (response.isSuccessful()) {
                     try {
@@ -482,7 +496,7 @@ public class AiService {
                 } else if (response.code() == 503 && attempt < 3) {
                     long delayMs = (attempt + 1) * 1500L;
                     new android.os.Handler(android.os.Looper.getMainLooper()).postDelayed(
-                            () -> generateChatResponseWithRetry(history, callback, attempt + 1),
+                            () -> generateChatResponseWithRetry(history, callback, attempt + 1, currentCall),
                             delayMs
                     );
                 } else {
--- a/app/src/main/java/io/actifit/fitnesstracker/actifitfitnesstracker/PostSteemitActivity.java
+++ b/app/src/main/java/io/actifit/fitnesstracker/actifitfitnesstracker/PostSteemitActivity.java
@@ -108,6 +108,7 @@ import kotlinx.coroutines.CoroutineStart;
 import kotlinx.coroutines.Dispatchers;
 import android.widget.ProgressBar;
 import android.widget.ImageButton;
+import com.google.android.material.snackbar.Snackbar;
 import android.widget.Spinner;
 import android.widget.ArrayAdapter;
 
@@ -2145,7 +2146,6 @@ public class PostSteemitActivity extends BaseActivity implements View.OnClickLis
     }
 
     private List<ChatMessage> aiChatHistory = null;
-    private ChatAdapter aiChatAdapter = null;
     private void showAiPopup() {
         AlertDialog.Builder builder = new AlertDialog.Builder(this);
         View dialogView = getLayoutInflater().inflate(R.layout.ai_popup, null);
@@ -2163,12 +2163,9 @@ public class PostSteemitActivity extends BaseActivity implements View.OnClickLis
         if (aiChatHistory == null) {
             aiChatHistory = new ArrayList<>();
         }
-        if (aiChatAdapter == null) {
-            aiChatAdapter = new ChatAdapter(aiChatHistory);
-        }
+        ChatAdapter chatAdapter = new ChatAdapter(aiChatHistory);
         chatRecyclerView.setLayoutManager(new LinearLayoutManager(this));
-        chatRecyclerView.setAdapter(aiChatAdapter);
-
+        chatRecyclerView.setAdapter(chatAdapter);
 
         boolean hasAiResponse = false;
         for (int i = aiChatHistory.size() - 1; i >= 0; i--) {
@@ -2186,14 +2183,19 @@ public class PostSteemitActivity extends BaseActivity implements View.OnClickLis
         }
         dialog.show();
 
+        // Holds the last user message so Retry can resend it even though the
+        // input field has already been cleared by the time a failure comes back.
+        final String[] lastUserMessage = {null};
+
         btnQuery.setOnClickListener(v -> {
             String userText = aiInputText.getText().toString().trim();
             if (userText.isEmpty()) {
                 return;
             }
 
+            lastUserMessage[0] = userText;
             aiChatHistory.add(new ChatMessage(ChatMessage.ROLE_USER, userText));
-            aiChatAdapter.notifyItemInserted(aiChatHistory.size() - 1);
+            chatAdapter.notifyItemInserted(aiChatHistory.size() - 1);
             chatRecyclerView.scrollToPosition(aiChatHistory.size() - 1);
             aiInputText.setText("");
 
@@ -2209,7 +2211,7 @@ public class PostSteemitActivity extends BaseActivity implements View.OnClickLis
                         loadingIndicator.setVisibility(View.GONE);
                         btnQuery.setEnabled(true);
                         aiChatHistory.add(new ChatMessage(ChatMessage.ROLE_AI, result));
-                        aiChatAdapter.notifyItemInserted(aiChatHistory.size() - 1);
+                        chatAdapter.notifyItemInserted(aiChatHistory.size() - 1);
                         chatRecyclerView.scrollToPosition(aiChatHistory.size() - 1);
                         if (!result.isEmpty()) {
                             btnAccept.setEnabled(true);
@@ -2224,19 +2226,37 @@ public class PostSteemitActivity extends BaseActivity implements View.OnClickLis
                         if (isFinishing() || isDestroyed()) return;
                         loadingIndicator.setVisibility(View.GONE);
                         btnQuery.setEnabled(true);
-                        aiChatHistory.add(new ChatMessage(ChatMessage.ROLE_AI, errorMessage));
-                        aiChatAdapter.notifyItemInserted(aiChatHistory.size() - 1);
-                        chatRecyclerView.scrollToPosition(aiChatHistory.size() - 1);
+                        // Don't add errors to chat history: keeps a stale error from being
+                        // inserted into the post later, and keeps errors out of what gets
+                        // replayed to the model on subsequent requests.
+                        Snackbar.make(dialogView, errorMessage, Snackbar.LENGTH_LONG)
+                                .setAction(R.string.retry_action, retryView -> {
+                                    if (lastUserMessage[0] != null) {
+                                        aiInputText.setText(lastUserMessage[0]);
+                                        btnQuery.performClick();
+                                    }
+                                })
+                                .show();
                     });
                 }
             });
         });
 
         btnAccept.setOnClickListener(v -> {
-            // Apply the most recent AI reply to the post
+            // Insert the most recent AI reply at the current cursor position
+            // (or at the end if the field was never focused / has no selection).
             for (int i = aiChatHistory.size() - 1; i >= 0; i--) {
                 if (!aiChatHistory.get(i).isUser()) {
-                    steemitPostContent.setText(aiChatHistory.get(i).getText());
+                    String aiText = aiChatHistory.get(i).getText();
+                    Editable editable = steemitPostContent.getText();
+                    int start = steemitPostContent.getSelectionStart();
+                    int end = steemitPostContent.getSelectionEnd();
+                    if (start < 0 || end < 0) {
+                        start = editable.length();
+                        end = editable.length();
+                    }
+                    editable.replace(start, end, aiText);
+                    steemitPostContent.setSelection(start + aiText.length());
                     break;
                 }
             }
--- a/app/src/main/java/io/actifit/fitnesstracker/actifitfitnesstracker/PostSteemitActivity.java
+++ b/app/src/main/java/io/actifit/fitnesstracker/actifitfitnesstracker/PostSteemitActivity.java
@@ -2145,7 +2145,9 @@ public class PostSteemitActivity extends BaseActivity implements View.OnClickLis
                 });
     }
 
-    private List<ChatMessage> aiChatHistory = null;
+    private List<ChatMessage> aiChatHistory = new ArrayList<>();
+
+    private boolean aiRequestInFlight = false;
     private void showAiPopup() {
         AlertDialog.Builder builder = new AlertDialog.Builder(this);
         View dialogView = getLayoutInflater().inflate(R.layout.ai_popup, null);
@@ -2160,9 +2162,6 @@ public class PostSteemitActivity extends BaseActivity implements View.OnClickLis
 
         EditText steemitPostContent = findViewById(R.id.steemit_post_text);
 
-        if (aiChatHistory == null) {
-            aiChatHistory = new ArrayList<>();
-        }
         ChatAdapter chatAdapter = new ChatAdapter(aiChatHistory);
         chatRecyclerView.setLayoutManager(new LinearLayoutManager(this));
         chatRecyclerView.setAdapter(chatAdapter);
@@ -2183,12 +2182,17 @@ public class PostSteemitActivity extends BaseActivity implements View.OnClickLis
         }
         dialog.show();
 
-        // Holds the last user message so Retry can resend it even though the
-        // input field has already been cleared by the time a failure comes back.
-        //final String[] lastUserMessage = {null};
+        // Tracks whether this specific popup instance is still on screen, so a
+        // response that arrives after the user closed it doesn't touch the now
+        // -detached adapter/RecyclerView (they'd be replaced by fresh ones on
+        // the next open anyway).
+        final boolean[] popupOpen = {true};
+        dialog.setOnDismissListener(d -> popupOpen[0] = false);
 
         Runnable[] sendRequest = new Runnable[1];
         sendRequest[0] = () -> {
+            if (aiRequestInFlight) return;
+            aiRequestInFlight = true;
             loadingIndicator.setVisibility(View.VISIBLE);
             btnQuery.setEnabled(false);
 
@@ -2198,12 +2202,16 @@ public class PostSteemitActivity extends BaseActivity implements View.OnClickLis
                 public void onSuccess(String result) {
                     runOnUiThread(() -> {
                         if (isFinishing() || isDestroyed()) return;
+                        aiRequestInFlight = false;
+                        if (!result.isEmpty()) {
+                            aiChatHistory.add(new ChatMessage(ChatMessage.ROLE_AI, result));
+                        }
+                        if (!popupOpen[0]) return;
                         loadingIndicator.setVisibility(View.GONE);
                         btnQuery.setEnabled(true);
-                        aiChatHistory.add(new ChatMessage(ChatMessage.ROLE_AI, result));
-                        chatAdapter.notifyItemInserted(aiChatHistory.size() - 1);
-                        chatRecyclerView.scrollToPosition(aiChatHistory.size() - 1);
                         if (!result.isEmpty()) {
+                            chatAdapter.notifyItemInserted(aiChatHistory.size() - 1);
+                            chatRecyclerView.scrollToPosition(aiChatHistory.size() - 1);
                             btnAccept.setEnabled(true);
                             btnAccept.setAlpha(1f);
                         }
@@ -2214,14 +2222,33 @@ public class PostSteemitActivity extends BaseActivity implements View.OnClickLis
                 public void onFailure(String errorMessage) {
                     runOnUiThread(() -> {
                         if (isFinishing() || isDestroyed()) return;
+                        aiRequestInFlight = false;
+
+                        if (!popupOpen[0]) {
+                            if (!aiChatHistory.isEmpty() && aiChatHistory.get(aiChatHistory.size() - 1).isUser()) {
+                                aiChatHistory.remove(aiChatHistory.size() - 1);
+                            }
+                            return;
+                        }
+
                         loadingIndicator.setVisibility(View.GONE);
                         btnQuery.setEnabled(true);
-                        // Don't add errors to chat history: keeps a stale error from being
-                        // inserted into the post later, and keeps errors out of what gets
-                        // replayed to the model on subsequent requests.
-                        Snackbar.make(dialogView, errorMessage, Snackbar.LENGTH_LONG)
-                                .setAction(R.string.retry_action, retryView -> sendRequest[0].run())
-                                .show();
+                        Snackbar snackbar = Snackbar.make(dialogView, errorMessage, Snackbar.LENGTH_LONG)
+                                .setAction(R.string.retry_action, retryView -> sendRequest[0].run());
+                        snackbar.addCallback(new Snackbar.Callback() {
+                            @Override
+                            public void onDismissed(Snackbar transientBottomBar, int event) {
+                                if (event == DISMISS_EVENT_ACTION) return;
+                                if (!aiChatHistory.isEmpty() && aiChatHistory.get(aiChatHistory.size() - 1).isUser()) {
+                                    int removedIndex = aiChatHistory.size() - 1;
+                                    aiChatHistory.remove(removedIndex);
+                                    if (popupOpen[0]) {
+                                        chatAdapter.notifyItemRemoved(removedIndex);
+                                    }
+                                }
+                            }
+                        });
+                        snackbar.show();
                     });
                 }
             });
@@ -2242,10 +2269,8 @@ public class PostSteemitActivity extends BaseActivity implements View.OnClickLis
         });
 
         btnAccept.setOnClickListener(v -> {
-            // Insert the most recent AI reply at the current cursor position
-            // (or at the end if the field was never focused / has no selection).
             for (int i = aiChatHistory.size() - 1; i >= 0; i--) {
-                if (!aiChatHistory.get(i).isUser()) {
+                if (!aiChatHistory.get(i).isUser() && !aiChatHistory.get(i).getText().isEmpty()) {
                     String aiText = aiChatHistory.get(i).getText();
                     Editable editable = steemitPostContent.getText();
                     int start = steemitPostContent.getSelectionStart();
--- a/app/src/main/java/io/actifit/fitnesstracker/actifitfitnesstracker/PostSteemitActivity.java
+++ b/app/src/main/java/io/actifit/fitnesstracker/actifitfitnesstracker/PostSteemitActivity.java
@@ -2144,6 +2144,8 @@ public class PostSteemitActivity extends BaseActivity implements View.OnClickLis
                 });
     }
 
+    private List<ChatMessage> aiChatHistory = null;
+    private ChatAdapter aiChatAdapter = null;
     private void showAiPopup() {
         AlertDialog.Builder builder = new AlertDialog.Builder(this);
         View dialogView = getLayoutInflater().inflate(R.layout.ai_popup, null);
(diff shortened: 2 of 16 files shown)


mcfarhat:
@AsmaaZeid nice, focused work — you found the actual root cause (history and adapter were locals inside `showAiPopup()`, so every open rebuilt them) rather than papering over the symptom, kept it to ~45 lines in one method with zero scope creep, preserved the `isFinishing()/isDestroyed()` guards, and cleaned up the dead commented-out block on the way past. I also checked all 14 `values-*/strings.xml`: every one parses cleanly, exactly one `</resources>`, **zero duplicate keys**, and `ai_accept_button` is defined in all of them — no locale left behind. Given that XML integrity has broken the build here before, that's worth calling out.

Also good: "persist session" is **RAM-only**. No chat content touches disk, so there's no health-data-at-rest question, nothing to clear on logout, and no serialisation to get wrong. Right default.

Three things to fix before merge. All three come from one subtle thing — **moving state from dialog scope to activity scope changes what "fresh" means** — so they're related, not three separate mistakes.

**1 + 2. Errors shouldn't go into the chat history at all.**

Right now a failure is appended as a `ROLE_AI` message (`PostSteemitActivity:2227`), which makes it indistinguishable from a real reply. Two consequences, both new because history now survives closing the popup:

- **A user can publish a raw API error as their post.** Fail a request → close → reopen: your restore-state loop (`:2173-2182`) only checks `!isUser() && !getText().isEmpty()`, so an error string passes and **Insert is enabled**. Tapping it runs `:2237-2239`, which walks back to the last non-user message and does `steemitPostContent.setText("API error: 429 - {\"error\":{…}}")` — silently replacing the post body with a Gemini error payload.
- **Errors poison the model context permanently.** `AiService.generateChatResponseWithRetry` serialises *every* entry in `history`, so each failure leaves a `"role":"model"` turn containing the error text that is re-sent with every subsequent request for the life of the activity. Previously closing the popup flushed it.

**The fix we'd like: don't add errors to `aiChatHistory` at all — surface them as a Snackbar (ideally with a "Retry" action) instead.** An error isn't part of the conversation, and putting it in the conversation list is what created both bugs. This fixes them *by construction* — if the error never enters the list, it can't be inserted and can't be replayed — with no extra flags or guards to remember later. Smallest diff and the hardest to regress.

**3. Don't keep the adapter as a field — keep only the list.**

`RecyclerView.setAdapter()` registers an observer that's only unregistered when *that* RecyclerView swaps adapters; `onDetachedFromWindow` doesn't do it. Each `showAiPopup()` inflates a fresh `ai_popup` with a fresh RecyclerView that registers on the shared `aiChatAdapter` and is never released — so N open/close cycles retain N dialog view hierarchies plus their ViewHolders until the activity dies, and every `notifyItemInserted` fans out to all N stale observers.

Fix is *less* code than what's there: keep `aiChatHistory` as the field, and do `new ChatAdapter(aiChatHistory)` on each open. The conversation still persists (it lives in the list), and the leak disappears.

**Follow-ups — not blocking this PR, no need to do them here:**
- Closing the popup doesn't cancel the in-flight OkHttp call, and the callback still holds the *old* dialog's views. Reopen mid-flight and the send button is enabled (so a second concurrent request can fire), the new spinner never shows, and the reply enables the *old* Insert button. A `private boolean aiRequestInFlight` applied on open would cover it.
- History is unbounded — the full transcript goes out on every request, so size, latency and free-tier quota grow monotonically. Worth capping to the last ~20 turns.
- Rotating with the popup open still throws `WindowLeaked` (`onDestroy` never dismisses the dialog) — pre-existing from #73.
- Small one: the Arabic string renders `ادراج`; correct orthography is `إدراج` (alef with hamza below).

**One open question, not a change request:** now that the button says "Insert", it's more visible that `:2239` **replaces** the entire post body rather than inserting at the cursor — ignoring existing content, with no undo. We're deciding whether it should insert at the cursor or keep replace-with-confirmation; I'll come back to you on it, so leave it as-is for now.

Fix the three above and I'll merge — this is good work.

mcfarhat:
@AsmaaZeid follow-up on the open question from my last comment — **decision: Insert should insert at the cursor**, not replace the post body. So please include this in the same round as the other three fixes.

Concretely, at `PostSteemitActivity:~2239`, instead of `steemitPostContent.setText(aiText)`:

```java
int start = Math.max(steemitPostContent.getSelectionStart(), 0);
int end   = Math.max(steemitPostContent.getSelectionEnd(), 0);
steemitPostContent.getText().replace(Math.min(start, end), Math.max(start, end), aiText);
```

`Editable.replace()` handles all three cases in one call: no selection (start == end) inserts at the caret, an active selection is replaced by the AI text, and an empty body just appends. It also keeps the existing content and, because it edits the `Editable` rather than swapping it, the user can still undo.

Two details worth getting right:
- **If the field was never focused**, `getSelectionStart()` returns -1 — the `Math.max(..., 0)` guards handle that, but it means the text lands at position 0 rather than the end. If you'd rather it append in that case, fall back to `steemitPostContent.length()` when the selection is -1.
- **Move the caret after the inserted text** so the user can keep typing: `steemitPostContent.setSelection(Math.min(start, end) + aiText.length())`.

Worth a quick manual check of the four cases: empty body, caret mid-text, text selected, and field never focused.

Everything else from my previous comment stands (errors → Snackbar rather than chat history, and don't retain the adapter). Ping me when it's pushed.

AsmaaZeid:
Pushed the fixes from your review:

1&2. Errors are no longer added to chat history — shown as a Snackbar with Retry, which resends the actual last message (not a click on the now-cleared input field)
3. Removed the aiChatAdapter field — adapter is now local and rebuilt each time the popup opens
4. Insert now inserts at the cursor position (using Editable.replace()), with a fallback to end-of-text when there's no active selection, and moves the caret after the inserted text
5. Fixed the Arabic spelling of the Insert button label

Confirmed via gradlew assembleDebug — build succeeds. Haven't been able to test live on-device this round (no phone access at the moment), so flagging that in case you want to verify the cursor-insert and Retry behavior before merging.

mcfarhat:
@AsmaaZeid the three main items are done properly rather than superficially, and that distinction matters:

- **Errors are excluded by construction** (`:2224-2241`) — `onFailure` no longer touches `aiChatHistory` at all, so Insert-after-failure can't re-enable (the restore loop finds no non-user entry) and nothing error-shaped reaches Gemini. That's the right shape: no flag to remember, nothing to filter later. The comment explaining *why* is good too.
- **Adapter removal is genuinely less code** — `grep aiChatAdapter` returns nothing repo-wide; it's a local built per open. The observer leak is gone.
- **`Editable.replace()` preserves the TextWatcher chain** — the Markwon preview, the draft save to SharedPreferences, the char counter and the progress bar all stay in sync. A `setText()`-based fix would have quietly broken those, and you avoided it.
- Localising `retry_action` across **all 14** files first time, with correct translations and no XML breakage, in a repo where that has broken the build before — exactly right. I parsed all 14: one `</resources>` each, zero duplicate keys.
- And fixing the Arabic `ادراج` → `إدراج` (U+0625) — verified at codepoint level.

Also: flagging honestly that you couldn't device-test this round was the right call. The two blockers below are precisely what a manual pass would have caught.

**Three small things:**

**1. Retry double-adds the user turn.** `:2196-2197` stores `lastUserMessage[0]` **and** appends the user turn to `aiChatHistory` *before* the request. On failure that turn isn't rolled back, so Retry (`:2233-2237`) sets the input and calls `performClick()`, which appends it again:

```
send "summarize my run"  → history = [user:"summarize my run"]  → 429
tap Retry                → history = [user:"...", user:"..."]
```

It renders twice in the chat and **both** copies go into the next request (two consecutive `role:"user"` turns — wasted tokens and non-alternating history). Retry twice → three copies. One-line fix: in `onFailure`, `aiChatHistory.remove(aiChatHistory.size()-1)` before showing the Snackbar, so Retry re-adds it cleanly. Or route Retry through a shared `sendMessage(String)` that skips the append.

*(Good news on the sub-cases: empty resend is impossible — `lastUserMessage[0]` is only assigned from a non-empty trimmed string — and double-send is blocked because the Snackbar dismisses on action tap and `btnQuery` is disabled during a request.)*

**2. The `Math.min`/`Math.max` normalisation is missing.** You replaced it with a `start < 0 || end < 0` check, which isn't equivalent. `getSelectionStart() > getSelectionEnd()` is legal for a **backwards selection** (Shift+Left on a hardware/Bluetooth keyboard or Chromebook), and `SpannableStringBuilder.replace()` throws `IndexOutOfBoundsException("… has end before start")` — **crashing the activity on Insert**. Keep your `-1` fallback, just add:

```java
int lo = Math.min(start, end), hi = Math.max(start, end);
editable.replace(lo, hi, aiText);
steemitPostContent.setSelection(lo + aiText.length());
```

**3. Same hamza, new string.** `values-ar/strings.xml` `retry_action` is `اعادة المحاولة` — bare alef again. Should be `إعادة المحاولة` (U+0625), the identical fix you just made two lines above.

**Nits, optional:** the Snackbar is anchored to `dialogView`, so if the user closes the popup mid-request and it then fails, the Snackbar is never laid out and they get **no** feedback at all — anchoring to the activity's `android.R.id.content` would survive dismissal. The error text is the raw `"API error: " + code + " - " + responseBody`, which in a 2-line Snackbar is unreadable and is untranslated for 13 locales — better to show a localised generic string and `Log.e` the detail. And `:2249` matches on `!isUser()` while the restore loop at `:2172` also requires `!getText().isEmpty()` — worth aligning.

**One follow-up I'd now promote:** the in-flight request still isn't cancelled on close. With the adapter no longer shared, close-mid-request → reopen → the old callback fires `notifyItemInserted` on the *old* adapter while the new one's cached count is stale — that's the classic setup for `IndexOutOfBoundsException: Inconsistency detected`. Not caused by this change (the adapter fix was right), but it makes the `aiRequestInFlight` guard the natural next thing rather than a someday item.

Fix those three and I'll merge.

AsmaaZeid:
Pushed the fixes:

1. Retry no longer re-adds the user message — extracted the actual API call into a reusable sendRequest, so btnQuery adds the message once, and Retry just re-invokes sendRequest directly against the existing history (no duplication, no double-send to Gemini)
2. Normalized backwards selections with Math.min/Math.max before calling Editable.replace() — prevents the crash on a right-to-left drag selection
3. Fixed the same hamza issue in the Arabic retry_action string (اعادة -> إعادة)

Also double-checked the insert-at-cursor logic against your exact spec from the earlier comment — confirmed it already handles all three cases correctly (caret insert, selection replace, append when never focused).

Build confirmed via gradlew assembleDebug.

mcfarhat:
@AsmaaZeid re-reviewed at `6526fea`. **All three items are properly fixed**, and I verified each in the code rather than going by the commit message.

**A. Retry no longer duplicates the user turn.** The refactor is the right shape: `sendRequest[0]` holds only the API call, the append happens once in `btnQuery`'s listener, and `onFailure` calls `sendRequest[0].run()` directly — so retry re-sends the existing history with zero re-adds. Traced the failure path: history ends at `[user:"A"]` and retry sends exactly that.

**B. Selection normalisation — done, and every call site is covered.** `:2251-2262` keeps the `-1` fallback, then applies `Math.min`/`Math.max` to both bounds before `editable.replace(start, end, aiText)` and moves the caret to `start + aiText.length()`. I grepped the whole repo: the only other `getSelectionStart()`/`getSelectionEnd()` pair is in `Utils.java:484` (image upload) and it already used the correct pattern. Nothing unguarded remains.

**C. The Arabic hamza** is fixed, and all 14 locale files still parse with `retry_action` and `ai_accept_button` present in each.

## One thing I flagged as a nit last round that I want to promote — it can now destroy the user's text

`:2203` adds the AI message **unconditionally**, including when `result` is empty; the `!result.isEmpty()` guard at `:2206` only gates *enabling* the button. Meanwhile Insert at `:2247` picks the newest message matching `!isUser()` alone.

So: you get a good reply (Insert becomes enabled, and it's never disabled again) → you send again → Gemini returns an empty string → you select a paragraph in your post and tap Insert → it finds that empty AI message and runs `editable.replace(start, end, "")`, which **deletes the selected paragraph and inserts nothing**.

Under the old `setText()` behaviour this was cosmetic. Now that Insert does a real `replace()` over a selection, it's data loss. Two-line fix: don't add empty AI messages at `:2203`, and match `!isUser() && !getText().isEmpty()` at `:2247` so it agrees with the restore loop at `:2172`.

## And a correction to something I told you last round

I said double-send was blocked because `btnQuery` is disabled during a request. **That was wrong** — the Retry action doesn't go through the button at all, it calls `sendRequest[0].run()` directly, so nothing stops it. Concretely: send "A" → it fails and the Snackbar appears → within its ~2.75s you type "B" and send (now in flight) → you tap Retry → two concurrent requests, doubled quota, both replies appended, and whichever finishes first hides the spinner while the other is still running.

An `aiRequestInFlight` boolean checked at the top of `sendRequest` closes this *and* the in-flight follow-up below, in one line.

## Should-fix

**An un-retried failure leaves an orphan user turn.** `btnQuery` always appends, `onFailure` never rolls back, and nothing prunes. Miss the Snackbar and your next message produces `[user:"A", user:"B"]` — two consecutive `user` turns replayed to Gemini forever, with a user bubble in the chat that has no reply and no error marker. It accumulates. This is inherent to the option you picked (keeping errors out of history), and the clean version is either rolling back on failure with Retry re-adding, or a `Snackbar.Callback` that rolls back if it's dismissed without action.

**In-flight requests still aren't cancelled when the popup closes** — this is the follow-up I promoted last round and it's the highest-severity thing left. The callbacks check `isFinishing()/isDestroyed()` but not whether the dialog is still showing. Send → close the popup mid-request → reopen (which builds a fresh adapter over the same list) → the old callback appends to the history and notifies the **old** adapter. The new popup shows nothing and Insert stays greyed out. The next send then fires `notifyItemInserted` against a RecyclerView whose cached count is stale, which is the classic `IndexOutOfBoundsException: Inconsistency detected`. The desync I've verified from the code; the exception is the usual consequence rather than something I reproduced.

## Nits

Delete the commented-out `//final String[] lastUserMessage = {null};` at `:2186-2188` and its now-inaccurate comment — you cleaned up exactly this kind of block last round. Retry now re-sends *current* history rather than the failed message, so if a send succeeds inside the Snackbar window, Retry appends an unprompted AI turn — narrow, no crash. `aiChatHistory` is declared inline at `:2148` and lazily initialised; `= new ArrayList<>()` with the other fields would remove the null check. And unchanged from last round: the Snackbar is still anchored to `dialogView` so a failure arriving after close shows on a detached window (no crash, but no feedback either), the error text is still the raw `"API error: " + code + " - " + body` and untranslated for 13 locales, history is still unbounded, and `onDestroy` still doesn't dismiss the dialog.

## Worth saying plainly

This repo has **no CI of any kind** — no `.github/` directory, and the head commit has zero check-runs. `app/src/test/` contains only the generated `assertEquals(4, 2+2)` scaffold. So `gradlew assembleDebug` is the entire verification loop, and that only proves it compiles.

Both the selection-normalisation logic and the Retry state machine are pure enough to unit-test cheaply, and both are exactly the kind of thing that breaks silently on the next change — the empty-message bug above is a case in point. Worth a handful of JUnit tests even without a pipeline to run them.

Fix the empty-message path and add the in-flight guard, and this is done.

AsmaaZeid:
Pushed a round that covers everything from your last review, including the "should-fix" and highest-severity items:

1. Empty AI responses no longer stored or insertable — the add-to-history block is now wrapped in !result.isEmpty(), and the Insert search loop double-checks !getText().isEmpty() as a second safety net.
2. In-flight request guard (aiRequestInFlight) prevents concurrent duplicate sends — checked at the top of sendRequest, reset on both success and failure.
3. Orphaned turn rollback — added a Snackbar.Callback; if the failure Snackbar is dismissed without tapping Retry (timeout or swipe), the pending user turn is removed from history instead of accumulating forever.
4. Popup-close-mid-request guard — added a popupOpen[0] tracker via dialog.setOnDismissListener. If a response arrives after the popup closed, history still gets updated (so it's there on reopen), but the adapter/RecyclerView are never touched — avoiding the desync/crash.
5. Nits — removed the dead lastUserMessage code, aiChatHistory now initialized inline.

This round I actually tested live on-device rather than just compiling, since you'd flagged there's no CI here:
- Selection-replace and cursor-insert-without-selection: correct
- Multiple sequential inserts: no duplication
- Backwards (right-to-left) selection: no crash
- Session persists correctly across close/reopen
- Failed message + Snackbar dismissed without Retry: message correctly disappears from history, doesn't get resent
- Send message → close popup immediately while loading → reopen: no crash, response correctly appears once it arrives

Didn't manage to cleanly reproduce the concurrent-double-request race live (hard to time), but traced it through the code and it's guarded structurally.

Ready for re-review whenever you get a chance.

mcfarhat (changes requested):
Reviewed at `e3d20392`. Two independent passes, and I verified the two load-bearing claims against the source myself.

**Three of your five items are properly closed, and you fixed something on `develop` that nobody had flagged.** The two that remain are the *same root cause*, so this is one concept to land rather than three more rounds.

## The best change in this PR isn't in the title

Your old `onFailure` appended the error string to history as a `ROLE_AI` turn. That meant error text was replayed to Gemini as a model turn, corrupting the conversation — **and** it was selectable by Insert as "the most recent AI reply", which in the old `setText` implementation meant a user could replace their entire post body with `"Network error: timeout"`. Moving errors to a Snackbar kills both. That's a real bug on `develop` that wasn't in your brief, and you removed a data-loss path in the process.

The Insert rewrite is also a genuine improvement: `setText(whole body)` → selection-aware `replace` with cursor restoration. And I checked the selection handling hard, because it's destructive — no-selection (`-1`), reversed selection, focus lost, empty body, span boundaries. **All correctly handled**, and crucially you read `start`/`end` fresh at click time rather than capturing them when the popup opened, which avoids the whole stale-index class. Clamping to `editable.length()` would be defence-in-depth only; I found no reachable path where it fires, so I'm not asking for it.

## Verified closed

**Empty responses.** `:2206` `if (!result.isEmpty())` stops it entering history, `:2273` matches `!isUser() && !getText().isEmpty()`, and you aligned the restore scan at `:2171` too — all three predicates now read identically. Your "second safety net" description is accurate rather than overstated.

**The in-flight guard closes the double-send.** I traced your exact scenario: send A → fails (flag reset) → send B → tap Retry → `:2194` returns. And you got the detail that breaks this when done wrong: `:2194` returns *before* `:2195` assigns, so the flag can't leak and kill the feature permanently. You also correctly hold it across `AiService`'s 503 ladder, which is easy to get backwards.

**Threading is clean.** Callbacks wrapped in `runOnUiThread`, and the 503 retry hops via `Handler(Looper.getMainLooper())`, so the shared list is only ever touched on the main thread. `aiRequestInFlight` needs no synchronisation — adding `volatile` would be cargo cult.

**Localisation is clean.** One new key `retry_action`, present exactly once in `values/` plus all 13 translated locales, `ai_accept_button` retranslated Accept→Insert in all 14, referenced as `R.string.retry_action` rather than hardcoded. No format specifiers in either key, so the `%s` mismatch class is structurally absent. Arabic renders correctly, hamza and all.

---

## Blocking — one root cause, three symptoms

**`popupOpen` is scoped to a single dialog, but the state it guards is scoped to the Activity.**

`aiChatHistory` (`:2148`) and `aiRequestInFlight` (`:2150`) are Activity fields. `popupOpen` (`:2189`) is a fresh array per `showAiPopup()`. So your callbacks ask "is the popup open?" about the popup that *started* the request, not the one that's live now. And `ChatAdapter` holds the list **by reference** — I checked: `this.messages = messages` with `getItemCount()` returning `messages.size()` live. So suppressing the `notify` while still mutating the list is the desync, not a guard against it.

Three consequences, all reachable by closing the popup mid-request and reopening (one tap, and the window is long — 60s read timeout plus up to 4 attempts on 503):

**1. The user's message is silently swallowed.** Nothing at popup-open derives `btnQuery` or the spinner from `aiRequestInFlight`, so popup B looks idle. Then:

```java
aiChatHistory.add(new ChatMessage(ROLE_USER, userText));
chatAdapter.notifyItemInserted(...);
aiInputText.setText("");
sendRequest[0].run();      // -> if (aiRequestInFlight) return;
```

The bubble appears, the input clears, no spinner, no error — and nothing is sent. Their message is gone. **Check the guard before mutating history and clearing the input**, and drive the spinner and button from `aiRequestInFlight` at open time.

**2. The list shrinks or grows under a live adapter.** On failure, `:2229` removes the entry and returns with no notify against the *live* adapter. On success, `:2207` appends and `:2209` skips the notify — so the reply is in history but never rendered, and `btnAccept` stays disabled, which is the exact symptom this PR set out to fix. Mutation-without-notify is the documented precondition for `IndexOutOfBoundsException: Inconsistency detected`. To be precise: the state loss is certain, the crash is likely but depends on when the next layout pass lands.

**3. The rollback can delete the wrong turn.** After a failure `btnQuery` is re-enabled while the Snackbar is still up (~2.75s). Send a new prompt inside that window and history is `[u_failed, u_new]`; the Snackbar times out and removes `size()-1` — the live one. The user's in-flight message vanishes and the reply arrives with no matching question. Fix by keeping a reference to the pending `ChatMessage` and removing by identity, not by position.

**The shape of the fix**, which handles all three: hold the live popup's adapter and views in an Activity field (null when closed) and have callbacks resolve the *current* popup rather than a captured one. Better still, have `AiService` return the OkHttp `Call` so `setOnDismissListener` can cancel it — then closing the popup genuinely ends the request and all three symptoms disappear by construction. `generateChatResponse` returns `void` today and never exposes the `Call`, which is a fair reason you reached for a flag instead.

---

## Non-blocking

- **History is unbounded, and this PR makes it effectively worse.** It used to be a local created per popup open, so closing was a de-facto reset. Now it lives for the Activity's lifetime with no clear-conversation control, and the whole list is serialised into `contents` on every send — so tokens grow quadratically in a long session. Worth a turn cap plus a clear affordance.
- **An empty success leaves an orphan turn with zero feedback.** When `result` is `""`, nothing is added, the spinner hides, the button re-enables, and the user's question sits there with no reply and no error — and that orphan user turn is replayed forever. Routing empty results into the same Snackbar-plus-rollback path as a failure closes it.
- **Retry silently no-ops if a request is in flight**, and because the Snackbar then dismisses with `DISMISS_EVENT_ACTION`, the rollback is skipped too — so the orphan is retained and the user's tap did nothing visible.
- Error text is still raw and English-only in all 14 locales — `"API error: " + code + " - " + responseBody` dumps truncated Gemini JSON into a 2-line Snackbar. Worth mapping to localised strings.
- Insert has no undo, and the `TextWatcher` persists the post body to `SharedPreferences` on every change, so an unwanted Insert is committed to the draft immediately. Strictly better than before; a confirmation step or a distinct "Replace selection" vs "Append" affordance is worth considering.
- `steemitPostContent` is already an Activity field (`:130`); `:2163` shadows it with a redundant `findViewById`. And this round deleted the accurate comment above `btnAccept` explaining the cursor-insert semantics — that one was worth keeping.

## One build risk only you can check

`unofficial_strings.xml` lives in the same `values/` folder as `strings.xml` and is gitignored, so no reviewer can see it. **If it already declares `retry_action`, the official build fails on a duplicate resource.** Low likelihood, but please confirm against your local copy before this merges — I can't.

## On verification

Neither pass could build: `google-services.json` and `unofficial_strings.xml` are gitignored and absent from a public clone. So everything above is read from source, not observed on a device — I'd rather say that than imply otherwise. **Please exercise the close-mid-request-then-reopen ordering yourself**, since that's the one that matters and it's the one your own testing narrowly missed: you tested close → wait → reopen, where the request had already finished. The broken ordering is close → reopen *while still loading*.

That distinction is a test-design lesson rather than a diligence problem — and you did report your device testing specifically, including naming the case you couldn't reproduce instead of implying you had. That's exactly what I asked for last round, and it's why this review could focus on one concept instead of arguing about what was tested.

AsmaaZeid:
Pushed the architectural fix you suggested — AiService.generateChatResponse now returns a CancellableRequest wrapping the current OkHttp Call, updated across retries. Closing the popup (dialog.setOnDismissListener) genuinely cancels it now instead of just guarding around stale state.

Specifically:
1. Message-swallowed: btnQuery now checks aiRequestInFlight before accepting a new message, and closing/reopening mid-request cleanly resets state (see below) instead of leaving it looking idle while actually blocked.
2. List-desync: since the old request's callbacks now genuinely never fire once cancelled (isCanceled() guards in AiService, plus the Call is truly aborted), there's no more mutate-without-notify happening at all.
3. Wrong-turn rollback: the Snackbar-dismissed-without-retry path now removes by identity (pendingTurn reference + indexOf) instead of size()-1.

One thing I found testing this live that wasn't in your review: cancelling a request needs to reset aiRequestInFlight immediately at cancel time, not rely on the (now-skipped) onFailure callback — otherwise the flag gets stuck true forever after any cancellation, permanently blocking new messages. Also added: the orphaned user turn from a cancelled request gets removed from history on dismiss too, same as the existing unretried-failure rollback, so a cancelled question doesn't confuse the AI's context on the next message.

Tested live, specifically the close-mid-request-then-reopen scenario you flagged as untested last round:
- Close while loading, reopen immediately: popup now shows idle and accepts a new message right away (previously got permanently stuck)
- The cancelled message is removed from history, doesn't get sent as stale context to the next question
- Confirmed via gradlew assembleDebug — build succeeds

Regarding unofficial_strings.xml — I don't have that file locally (it's gitignored as you noted), so I can't confirm whether it declares retry_action. Could you check on your end, or let me know how you'd like me to verify?

mcfarhat (changes requested):
## Re-reviewed at `e3c317d1` — right architecture, one hole left

You took the cancellation fix rather than patching around it, and implemented it cleanly. Symptom 1 is closed outright; 2 and 3 are closed on the plain path and reopen through a single narrow gap. **One small commit from an approve.**

This is also the first round where I could actually build it: `./gradlew assembleDebug` → **BUILD SUCCESSFUL**, real 50 MB APK, resource merge passed. (For anyone doing this later — the `release` `signingConfig` block is evaluated at *configuration* time, so the build fails at configure without a `keystore.properties` and keystore, even for a debug build.)

### Symptom 1 — user's message silently swallowed: **CLOSED**

`:2270` — the `if (aiRequestInFlight) return;` guard is now the **first statement** in the `btnQuery` listener, ahead of the `add` / `notifyItemInserted` / `setText("")` sequence at `:2276-2279`. The append-then-bail ordering is gone. `:2179-2182` also drives the spinner and disables `btnQuery` from `aiRequestInFlight` at popup-open, so a popup can't look idle over a live request.

### Symptoms 2 and 3 — fixed on the plain path

`dialog.setOnDismissListener` (`:2194`) reaches `Call.cancel()` at `AiService.java:454`. OkHttp delivers `onFailure` with a `"Canceled"` `IOException`, and `AiService:482`'s `if (call.isCanceled()) return;` swallows it — no spurious network error and no double rollback, because the dismiss handler already rolled back at `:2199-2201`.

Symptom 3's positional `size()-1` removal is replaced by identity lookup (`pendingTurn` at `:2243`, `indexOf` at `:2255`). I checked that `ChatMessage` does **not** override `equals()`, so `indexOf` is true reference identity — which is exactly what this needs. Every mutation on that path is now paired with a notify.

---

## Blocking — the 503 retry ladder escapes cancellation

`AiService.java:449-457`:

```java
public CancellableRequest generateChatResponse(...) {
    final Call[] currentCall = new Call[1];
    generateChatResponseWithRetry(history, callback, 0, currentCall);
    return () -> { if (currentCall[0] != null) currentCall[0].cancel(); };
}
```

`cancel()` cancels whatever `Call` is *currently* in the array. But on a 503, `:496-500` schedules the next attempt on `Handler(Looper.getMainLooper()).postDelayed(...)` at 1500/3000/4500 ms. During that delay `currentCall[0]` holds a call that has **already completed** — cancelling it is a no-op. The delayed runnable then creates a brand-new `Call` at `:478-479` that nobody can cancel, because `activeRequest[0]` was nulled at `PostSteemitActivity:2197`.

Unprotected window: **~9 seconds** across the three retries. And it's exactly the situation the ladder exists for — a 503 is precisely when someone gets impatient and closes the popup.

**Reproduction** (Gemini returning 503, or stub it):

1. Open the popup, type "hello", send → `aiRequestInFlight = true`
2. 503 → retry #1 scheduled in 1.5s
3. Within that 1.5s, close the popup. Dismiss cancels an already-completed call (no-op), sets `aiRequestInFlight = false`, removes "hello"
4. 1.5s later the retry issues a **new, uncancelled** call
5. Reopen. Popup B looks idle — correctly, the flag is false. Type "world", send. B's adapter is now live on the shared `aiChatHistory`
6. The escaped retry succeeds. Popup **A**'s `onSuccess` (`:2216`) runs — `isFinishing()` is false — so it clears `aiRequestInFlight` (**clobbering popup B's in-flight flag**), appends to the shared history at `:2223`, and calls `notifyItemInserted` on **popup A's detached adapter** at `:2224`

That's symptom 2 verbatim. `ChatAdapter` still holds the list by reference (`this.messages = messages`, `getItemCount()` reads it live), so popup B's adapter sees its item count jump with no notify: `btnAccept` stays disabled despite a reply existing — the exact bug this PR set out to fix — and the next layout pass is an `IndexOutOfBoundsException: Inconsistency detected` candidate.

Being straight about the limits here: if the escaped retry *fails* instead, popup A's `onFailure` captures `pendingTurn` as popup B's live "world" turn and the Snackbar timeout would remove it. But that Snackbar is anchored to A's detached `dialogView`, and whether `onDismissed` fires on a never-attached Snackbar depends on `BaseTransientBottomBar` behaviour I did not exercise on a device. So treat the failure branch as *unproven*; the success branch above is deterministic and sufficient on its own.

**The fix is small** — cancellation needs to cover the ladder, not just the socket:

```java
public CancellableRequest generateChatResponse(...) {
    final Call[] currentCall = new Call[1];
    final boolean[] canceled = {false};
    generateChatResponseWithRetry(history, callback, 0, currentCall, canceled);
    return () -> { canceled[0] = true; if (currentCall[0] != null) currentCall[0].cancel(); };
}
```

Check it at the top of `generateChatResponseWithRetry` and inside the `postDelayed` lambda. Both are main-thread-only today, so a plain `boolean[]` is enough — no `volatile` needed.

---

## Non-blocking

- **Response body leak on the cancel path.** `AiService:486` returns from `onResponse` without `response.close()`. OkHttp requires the caller to close the body, so a cancelled-but-delivered response leaks a connection. One line.
- **TOCTOU on `isCanceled()`.** `:486` checks on the OkHttp thread, then `PostSteemitActivity:2216` `runOnUiThread` defers with no re-check. A dismiss landing in that window leaves an orphan AI reply whose question was already deleted. Milliseconds wide, so negligible alone — but re-checking `call.isCanceled()` inside the UI runnable closes it for free, and it's the same discipline as the fix above.
- **`pendingTurn` dropped the `.isUser()` check** (`:2243`). Rollback previously only removed a trailing *user* turn; now it removes whatever is last. Unreachable on the plain path, but reachable through the 503 hole, where it would delete an AI reply. Worth restoring as defence in depth.
- **`:2179-2182` is effectively dead and would brick popup B if it ever fired.** The dismiss handler always resets `aiRequestInFlight`, so it's never true at popup-open on any path I could reach — and if it ever became reachable, nothing in popup B re-enables `btnQuery`, because the callbacks that would are holding popup A's views. Not a bug today; flagging it because it looks load-bearing and isn't.
- **The user's typed text is lost on close.** The rollback deletes the pending turn and the input was cleared at `:2279`, so their question is unrecoverable. Consider restoring it to `aiInputText` on reopen.
- Carried over, still open: unbounded history with no clear-conversation control (quadratic token growth), empty-success leaving an orphan turn with no feedback (`:2222`), raw English `"API error: " + code` in all 14 locales, and Insert having no undo against the `TextWatcher` that persists to `SharedPreferences`.

## Verified not regressed

Empty-response guard identical across all three predicates (`:2168`, `:2222`, `:2288`). In-flight guard still has the `return` at `:2207` *before* the assignment at `:2208`, and Retry routes through `sendRequest[0]` so it's covered.

I specifically hunted for the stuck-`true` failure mode I said would be worse than the bug itself — a permanently dead feature — and **could not construct a path to it**. `:2198` clears the flag unconditionally whenever `activeRequest[0] != null`, and `activeRequest[0]` is assigned synchronously on the main thread at `:2213` before any callback can run. Even the escaped retry resets it.

Localisation: `retry_action` and `ai_accept_button` each present exactly 14 times, no format specifiers — confirmed by a real resource merge, not grep.

## Only you can check this one

`unofficial_strings.xml` is gitignored. I proved the `.example` template has **zero** name overlap with `values/strings.xml` (116 names vs 730), so the public build is clean — but if your real local copy declares `retry_action`, the official build fails on a duplicate resource. No reviewer can see that file.

## What you did well

Taking the architectural fix was the right call, and `CancellableRequest` is a good shape for it — a minimal one-method interface that doesn't leak OkHttp types into the Activity, wired to `setOnDismissListener` so it fires on back button, outside tap and close button alike rather than just the one button.

The detail I want to single out: you handled the OkHttp cancellation contract without being told the specifics. Adding `if (call.isCanceled()) return;` to **both** `onFailure` and `onResponse` is what most people miss — the naive version guards only `onFailure` and then shows the user a "Network error: Canceled" toast. You got both the spurious-error and double-rollback paths right first time.

You also picked up the two small carried-over notes without being chased, and your comments at `:2190-2192` and `:2251-2253` explain *why* rather than restating *what*, which is the difference between a comment that ages well and one that doesn't.

One function in `AiService`, and this is done.

AsmaaZeid:
Pushed the fix for the 503 retry ladder — extended cancellation to check a canceled[0] flag at the top of generateChatResponseWithRetry and inside the postDelayed retry lambda, not just cancelling the live Call. This closes the ~9-second unprotected window across the three retries where a cancel would previously be a no-op against an already-completed call, letting the escaped retry fire uncancelled.

Applied exactly as specified — canceled[0] checked at both points, passed through on the recursive retry call.

Confirmed via gradlew assembleDebug — build succeeds. This scenario (503 mid-retry-delay + close popup) isn't something I can reliably reproduce live without a way to force a 503 from Gemini, so this one's verified at the code level against your exact repro steps rather than live-tested like the previous round.

Ready for re-review.

mcfarhat (approved):
## Approved at `6d231ee0`

The blocker is closed, and closed exactly right. Nine lines, no collateral changes.

```
:451  final boolean[] canceled = {false};
:454  canceled[0] = true;              <- first statement in the canceller
:455  currentCall[0].cancel();
:462  if (canceled[0]) return;         <- top of generateChatResponseWithRetry
:501  if (canceled[0]) return;         <- inside the postDelayed lambda
```

The one at `:501` is the part that's easy to get wrong. A check only at the top of the retry method looks correct and silently leaves an already-queued runnable live — you put it in both places. You also got the write-before-cancel ordering right at `:454`/`:455` without my saying it was load-bearing. The ~9-second window across the three retries is gone: cancel during a delay sets the flag, the runnable returns, no new `Call` is created, no callback reaches the Activity.

I went back through the adversarial checks rather than trusting the diff:

- **Stuck `aiRequestInFlight = true` — no path exists.** This was the failure mode I said would be worse than the original bug, and the new cancel path is where it would live, since *no callback fires at all*. Only the dismiss handler can clear it, and `:2196-2198` do cancel / null / clear unconditionally inside one `if (activeRequest[0] != null)` block. `activeRequest[0]` is assigned synchronously on the main thread at `:2213`, so it's never null while a request is live, and both callbacks clear the flag at `:2218`/`:2236` *before* nulling the handle. There's no observable `activeRequest == null && inFlight == true` state.
- **Threading re-derived.** I told you a plain `boolean[]` was sufficient, so I re-checked rather than assume my own advice held. Every read and write of `canceled[0]` and `currentCall[0]` is on the main thread — initial call from `btnQuery`/Snackbar action, retries via `Handler(Looper.getMainLooper())`, canceller via `setOnDismissListener`. The `:497-505` scheduling runs on the OkHttp thread but only touches the thread-safe `Handler`. No `volatile` needed; the advice was right.
- **The dismiss path removes without a notify at `:2199-2201`, and that's safe** — AOSP's `Dialog.dismissDialog()` calls `removeViewImmediate(mDecor)` *before* `sendDismissMessage()`, so the listener runs on a later loop turn against a fully detached RecyclerView. No layout pass can observe it.
- Nothing regressed: empty-response guard identical at `:2168`/`:2222`/`:2288`, in-flight `return` still precedes the assignment, `ChatMessage` still has no `equals()` override so `indexOf` is true reference identity, and `retry_action`/`ai_accept_button` are 14/14 — confirmed via `aapt2 dump resources` on the built APK rather than grep.

`./gradlew assembleDebug` → **BUILD SUCCESSFUL**, real 50 MB APK.

### Two one-liners you didn't take — worth a follow-up commit

Neither blocks this merge, but one has changed status since I raised it:

**1. `.isUser()` on `pendingTurn` (`:2243`) is now a real guard, not defence in depth.** Last round I said it was unreachable except through the 503 hole. That hole is closed — but it's now reachable on the *plain* path, all inside one Snackbar's 2.75s:

request 1 fails (Snackbar 1, `pendingTurn` = U1) → you send U2 → request 2 **succeeds**, history is `[U1, U2, A2]` → you tap **Retry on Snackbar 1** (that's `DISMISS_EVENT_ACTION`, so U1 isn't rolled back) → request 3 fails → `:2243` captures `pendingTurn` = **A2, an AI turn** → Snackbar 3 times out and deletes a valid AI reply.

A three-step human race against a full Gemini round-trip, so genuinely unlikely — but it's a one-word change and it now prevents something real.

**2. `response.close()` on the cancel path.** `:487` still returns without closing, and it's the only leak — the 503 and error branches are fine because `ResponseBody.string()` closes the source itself, and `.string()` is never reached at `:487`.

### Settled, for the record

The Snackbar-on-detached-view branch I flagged as *unproven* last round is now proven, and it's **benign**. From Material 1.12.0 source: `onDetachedFromWindow()` posts `onViewHidden(DISMISS_EVENT_MANUAL)` on the main-looper handler, which does dispatch `onDismissed` — so your rollback does fire, with `MANUAL` rather than `ACTION`, so `:2250` doesn't skip it. The detach runnable is posted during `removeViewImmediate`, i.e. before the dialog's own dismiss message, and both land before any human can reopen. History is consistent by the time popup B builds its adapter at `:2162`. No desync.

### Still only you can check

`unofficial_strings.xml` is gitignored. The template has 115 names against `values/strings.xml`'s 730 with **zero** overlap, so the public build is clean — but if your real local file declares `retry_action`, the official build fails on a duplicate resource.

Merging. Good work on this one — you took the architectural fix two rounds running rather than patching symptoms, and this last commit was exactly nine lines with no scope creep, which is the right size for where we are.

PR #88, opened 2026-08-09
Show gadget beneficiary name and avatar in market

Adam-Diab1:
## Summary

Shows the beneficiary username and profile picture by default on active Market gadget cards when the gadget is activated for a friend.

## Changes

- Preserve the saved `benefic` value from active gadget records
- Prefer active copies when multiple copies of the same gadget exist
- Show beneficiary username and Hive avatar on active friend gadgets
- Reset beneficiary/avatar state safely for recycled rows
- Update beneficiary display immediately after successful activation
- Refresh gadget state after deactivation to handle multiple active copies correctly
- Require explicit backend `status: "success"` before mutating local state
- Safely handle missing or malformed beneficiary data

## Verification

- `git diff --check` passes
- `assembleDebug` passes with Android Studio JDK 21
- Changes are limited to the Market gadget card flow


mcfarhat:
@Adam-Diab1 reviewed. The recycling discipline here is the best part — that's usually where this kind of change breaks, and you got it right. Two things to fix, both straightforward.

## What's solid

**`bindBeneficiaryIdentity` resets before it decides.** It calls `Glide.with(...).clear(avatar)`, sets the container `GONE`, blanks the username and restores `default_pic` *before* any early return. That's exactly right for a recycled `convertView` — the `Glide.clear()` in particular is the step people miss, and skipping it is what makes a recycled row briefly show the previous user's avatar. Same thinking behind the `friendBeneficiary.getTag()` check to clear a stale typed username when a row is reused for a different gadget.

**Defensive parsing of `benefic`.** `has()` → `isNull()` → `opt()` → `instanceof String` → `trim()` is the right amount of paranoia for a field coming off the API, and stripping a leading `@` means it renders correctly whether the backend stores `user` or `@user`.

**Prefer-active-copy.** Widening the condition to `|| (isActiveCopy && nonConsumedCopy != ACTIVECOPY)` so an active copy wins over an already-recorded bought copy is the correct fix for the multiple-copies case.

## Should fix

### 1. The `benefic` value goes on-chain through string concatenation

`MarketActivity.java`, the activation payload:

```java
cstm_params.put("json", "{\"transaction\": \"activate-gadget\" , \"gadget\": \"" + postEntry.id
        + "\" , \"benefic\": \"" + activationBeneficiary + "\"}");
```

`activationBeneficiary` is whatever the user typed. A `"` or `\` in that field produces malformed JSON, and a value like `x", "transaction": "something-else` injects keys into a `custom_json` we then **broadcast to the chain** — permanently, and signed by the user's posting key. This is pre-existing rather than introduced by you, but you're already validating this field two lines up, so it's the natural place to close it.

Two changes, both small:

```java
// build it as JSON rather than by concatenation
JSONObject payload = new JSONObject();
payload.put("transaction", "activate-gadget");
payload.put("gadget", postEntry.id);
if (postEntry.isFriendRewarding) payload.put("benefic", activationBeneficiary);
cstm_params.put("json", payload.toString());
```

and validate the username before broadcasting, since an invalid one wastes an on-chain op that can never succeed:

```java
if (!activationBeneficiary.matches("^[a-z][a-z0-9\\-.]{2,15}$")) { /* toast + return */ }
```

That also gives the user a clear error instead of a silent no-op when they type `@Bob` or a display name.

### 2. 62 lines of CRLF→LF conversion are mixed into the diff

A third of the `+173` is line-ending churn, not real change — whole blocks like the Volley imports and the `cstm_params` setup show as removed-and-re-added with identical content. Two problems: it hides the ~110 lines that actually matter, and it will conflict with anything else touching `MarketActivity`/`ProductAdapter` — Reem is working in the market screen right now.

Please either revert the EOL-only lines or split them into a separate "normalize line endings" commit. If we want LF everywhere as a rule, a `.gitattributes` with `*.java text eol=lf` is the durable fix and I'm happy to take that as its own PR.

## Nits

- `postEntry.id.equals(friendBeneficiary.getTag())` NPEs if `id` is ever null — `Objects.equals(postEntry.id, ...)` is free insurance.
- `Glide.with(getContext())` from an adapter throws `IllegalArgumentException` if the host activity is already destroyed. Rare, but this app has had lifecycle crashes before; worth an `isFinishing()/isDestroyed()` guard or passing a lifecycle-aware context.
- `while (normalizedUsername.startsWith("@"))` does the job; `replaceFirst("^@+", "")` says it in one line.

## One thing to confirm

You added `postEntry.remainingBoosts = postEntry.validityVal;` immediately before the `posts_consumed` recalculation. I read that as deliberate — now that the branch can run a second time for the same product (the prefer-active-copy change), it stops the previous copy's decrements carrying over. Just confirm that's the intent, and that with several non-consumed entries the value we end up showing is the active copy's rather than the last one iterated.

## Process

There's no Trello card linked to this PR, so it never showed up in my review queue — I only caught it on a manual sweep. Please link it to your existing card (or open one) so the board reflects it.

Fix the two above and I'll merge. Nice work on the recycling handling.

mcfarhat:
@Adam-Diab1 re-reviewed at `9773e757`. **The security item is properly fixed** — that was the one that mattered.

## 1. On-chain JSON injection — fixed, and fixed well

`ProductAdapter.java:693-699` builds the payload with `JSONObject` now, and `:685-691` validates before anything is broadcast. I attacked the regex rather than just eyeballing it:

| input | result |
|---|---|
| `x", "transaction": "something-else` | REJECT |
| `a"bc` / `a\bc` / embedded newline | REJECT |
| `@bob`, `Bob`, `1bob`, `bo b` | REJECT |
| `bob&x=1`, `bob#f`, `bob%26`, `bob/x` | REJECT |
| empty, 2 chars, 17 chars | REJECT |
| `bob`, `ac-ti.fit` | PASS |

I also checked no concatenated path survives with user input: the only remaining string-built payloads are `buy-gadget` (`:477`) and `deactivate-gadget` (`:901`), and both interpolate `postEntry.id` from the API, never the typed field. `R.string.invalid_username` already exists in 10 locales, so no missing-resource break.

**Worth knowing why that validation is load-bearing rather than cosmetic:** `:707` still builds the broadcast URL as `"&operation=[" + operation + "]"` with **no URL encoding**, and `buyUrl += "/" + activationBeneficiary`. Your regex is now the only thing keeping `&`, `#` and `%` out of that query string. Don't relax it without adding encoding first.

Two leftovers, both minor and one of them mine: the regex I handed you accepts `bob.`, `bob-` and `a..b`, which Hive itself rejects — that's my regex being loose, not your mistake, and the cost is a wasted on-chain op rather than a hole. And `@bob`/`Bob` now hard-reject, which is exactly what I asked for, but `bindBeneficiaryIdentity` already strips `@` for display — lowercasing and stripping `@` on input would be friendlier than rejecting.

## 2. Line-ending churn — not fixed

No `.gitattributes`, no normalization commit. And **I owe you a correction: I said 62 churn lines, the real number is 33.** I measured it properly this time:

| | added lines | churn (identical content) | of which CRLF→LF | real new |
|---|---|---|---|---|
| before | 173 | 39 | 38 | 134 |
| now | 187 | 34 | 33 | 153 |

The problem has actually got slightly worse in kind, if not in count: the files are now **mixed-EOL** — `ProductAdapter.java` has 1208 CRLF lines out of 1345, `MarketActivity.java` 460 of 475 — because you're writing LF lines into CRLF files. That's the worst of both, and it guarantees the conflict surface with Reem's market-screen work.

Either revert the EOL-only lines or add `.gitattributes` with `*.java text eol=lf` and normalize in one dedicated commit.

## 3. Nits

`Objects.equals` — **fixed** (`:105`, import at `:45`). Glide lifecycle guard — **fixed** (`:1313-1317`, and `minSdkVersion 28` so `isDestroyed()` is safe). `while (startsWith("@"))` — still there at `:1329`; purely cosmetic, take or leave.

Two new small ones: the validation at `:685` runs *after* `startAnimation` and after the auth `loginRequest` is already queued (`:653`), so an invalid username still costs a login round-trip — same pattern as the existing empty-check, but both belong above the network work. And `payload.put("gadget", postEntry.id)` would silently *drop* the key if `id` were ever null, since Android's `JSONObject.put` removes on null — unreachable in practice because `SingleProductModel` defaults it to `""`, but it sits oddly next to the `Objects.equals` you just added specifically to defend against a null `id`. Pick one belief about that field.

## 4. The question I asked

You didn't answer it, so I checked myself: the widened condition at `MarketActivity.java:210` means an active copy overrides a previously-recorded bought copy and, once `ACTIVECOPY` is set, no later entry overwrites it. So the displayed value **is** the first active copy's rather than the last iterated, and the `remainingBoosts = validityVal` reset is genuinely needed to stop the bought copy's baseline carrying into the active copy's decrement. **Your code is right.** Please still confirm it was deliberate, and consider a one-line comment — it's non-obvious enough that the next reader will wonder.

## Process

Two things: there's still **no Trello card linked** to this PR, which is why it never appeared in my review queue and I only caught it on a manual sweep. And you pushed a fix without replying to the review — a one-line "pushed, here's what I did" makes it much easier to know what to re-check.

One round left: split the EOL change out, confirm the boosts question, link the card. The nits are optional.

mcfarhat (changes requested):
Reviewed at `39626d79`. Two independent passes, plus my own check against the backend source where they disagreed.

Apologies for the wait on this one — you pushed on 08-09 and the review is three days late. That's a gap in how I sweep open PRs, not anything you did.

**One blocking item.** Everything else below is follow-up, and a fair amount of it is credit.

---

## The success-check change is correct — and it fixes a real bug

My two reviewers split on this, so I checked `actifit-bot` myself rather than pick a side. You changed `!response.has("error")` → `"success".equals(status)` at `:744` and `:941`, and left the two buy handlers at `:517` and `:1145` on `!has("error")`. One pass read that as an inconsistency worth blocking on. It isn't — it's exactly right:

```
app.js:8255  activateGadget       success -> {'status': 'success'}
                                  bad trx -> {status: 'error'}      <- no "error" key
app.js:8321  performDeactivateTrx success -> {'status': 'success'}
                                  bad trx -> {status: 'error'}      <- no "error" key
             buyGadget / buyGadgetHive     -> {'error': ...}         <- genuinely the error shape
```

So on `develop` today, a **failed** activation returns `{status:'error'}`, `!has("error")` is true, and the app shows the success toast and flips the row to active for an activation that never happened. Your change closes that. And you left the buy handlers alone because their endpoints really do use the `error` shape — you matched each endpoint's actual contract instead of blanket-refactoring all four. Most people would have "tidied" all of them and broken purchases. That's the single best judgement call in this PR.

## Your `remainingBoosts` explanation checks out

Verified against `MarketActivity.java:221-240`. The gate `NOCOPY || (isActiveCopy && nonConsumedCopy != ACTIVECOPY)` lets an active copy upgrade over a previously-selected bought copy; line 234 resets to `validityVal` before line 239 subtracts that copy's `posts_consumed`. No double-subtraction — once `ACTIVECOPY` is set the second clause can't re-fire, so multiple active copies can't stack. Your reasoning holds exactly as you described it.

---

## Blocking: deactivation can leave the row stuck showing "Deactivate"

`ProductAdapter.java:941-946`. You replaced the synchronous `postEntry.nonConsumedCopy = BOUGHTCOPY` with an async re-fetch, and `nonConsumedCopy` now only changes inside that callback at `:1282`. The error listener at `:1305` just logs.

```
user taps Deactivate on a flaky connection
  -> chain op broadcasts, deactivateGadget returns success, success toast shown
  -> the follow-up refresh GET times out
  -> row stays ACTIVECOPY, still renders the Deactivate button
  -> user taps Deactivate again on an already-deactivated gadget
  -> a second signed custom_json goes on-chain, and they get an error toast
```

To be clear, **re-fetching is the right design** — with multiple owned copies, deactivating one doesn't necessarily mean the row becomes "bought", and the old synchronous line was a guess. Only the failure path is missing. In the error listener, fall back to what the old code did:

```java
product.nonConsumedCopy = SingleProductModel.BOUGHTCOPY;
product.beneficiary = "";
product.remainingBoosts = product.validityVal;
notifyDataSetChanged();
```

Also drop the `notifyDataSetChanged()` at `:943` — it runs *before* the state changes, so it redraws the still-active layout.

---

## Follow-ups, not blocking

- **`remainingBoosts` jumps after deactivation.** `refreshProductStateAfterDeactivation` (`:1282`) sets `remainingBoosts = validityVal` and only subtracts `posts_consumed` when it finds an *active* copy. A deactivated copy keeps its consumed posts and `MarketActivity` does subtract them — so deactivate a gadget that used 2 of 5, the row shows 5, then re-enter the market and it shows 3.
- **The activate path asserts the beneficiary the client sent** (`:747`), not what the server confirmed, so if the backend normalises or rejects it the UI is wrong until reload. You already used the better pattern (re-fetch) on deactivate — worth applying it symmetrically.
- **`MarketActivity.java:221-222` changes behaviour on the 99% path.** The new `|| (isActiveCopy && nonConsumedCopy != ACTIVECOPY)` upgrades a row from bought to active when a later entry is active; previously the first match locked it in. It looks like a fix and the `validityVal` reset keeps it correct — but it's invisible in a PR titled "show beneficiary name and avatar". Please add it to the description and test a multi-copy gadget by hand.
- **RTL:** the new `LinearLayout` uses `paddingLeft`/`paddingRight`. With `supportsRtl="true"` and a `values-ar` locale, that mirrors wrong in Arabic — use `paddingStart`/`paddingEnd`.
- **`bindBeneficiaryIdentity` doesn't check `isFriendRewarding`** (`:1321`), so a stray server-side `benefic` on a normal gadget would render. One-word guard.
- **Display-path avatar URL isn't encoded** (`:1336`). Worth tightening for consistency, but I want to be precise since it would be easy to overstate: the substitution is in the path *after* the authority, so no value can change the host — no SSRF, no redirect. It's defence-in-depth, not a vulnerability.
- `MarketActivity.java:235` `getJSONArray`/`getInt` throw on a missing key, and the outer `catch(JSONException)` at `:352` doesn't call `progress.hide()` — one malformed entry spins the market forever. Pre-existing, but your change makes that block run more often. You used the `opt*` forms in your own new method; the two implementations just disagree.

---

## What you did well — specifically

**You defeated the view-recycling bug structurally rather than patching symptoms.** `bindBeneficiaryIdentity` is called unconditionally at `:109`, outside the inflate guard, and resets visibility, text and avatar *before* any early return. That is the number-one thing that kills features like this, and getting it right means no recycled row can ever flash the previous person's name or face — which here would have been a genuine, if cosmetic, privacy bug. No other adapter in this codebase does the `Glide.clear()` + placeholder properly; `VoteEntryAdapter.java:58` doesn't.

**You found and fixed a TOCTOU bug nobody asked about.** The old code re-read `friendBeneficiary.getText()` *inside* the async response handler, so if the row recycled between broadcasting the chain op and the callback, the confirmation URL carried a different beneficiary than the one actually signed on-chain. Hoisting it to a `final` local at `:677` makes the signed op and the API call provably agree. That's subtle and you caught it.

**The security fix is the right kind of fix.** You replaced string-concatenated JSON with a real `JSONObject`, which kills the injection class structurally rather than filtering for it, *and* you added the input guard — and the guard also closes the second path at `:733` where the beneficiary is concatenated into a URL. You covered both, not just the reported one.

Also: the `isFinishing()/isDestroyed()` Glide guard at `:1313`, which pre-empts the classic destroyed-activity crash; the EditText tag check at `:105-108`, which stops typed text bleeding into another product's row on recycle *and* preserves it across `notifyDataSetChanged()`; reusing the already-translated `invalid_username` instead of hardcoding English in a 15-locale app; `contentDescription` on the avatar; and a clean third commit — I diffed `9773e75..39626d7` with `-w` and the only non-whitespace change is one added comment.

## What neither pass could verify

Neither reviewer could build. `app/src/main/res/values/unofficial_strings.xml` and `app/google-services.json` are gitignored and absent from any clone, so `assembleDebug` is impossible for anyone without the private files — pre-existing and unrelated to this PR. Every resource you reference does exist (`invalid_username` at `strings.xml:316` and translated in `values-ar:298`, `hive_image_host_url`, `default_pic`, all three new layout IDs), and the changed files are syntactically clean, but **the layout was not seen rendered** — please check it yourself at a narrow width with a 16-character username before this lands.

Fix the deactivation error path and add the `MarketActivity` change to the description, and I'll merge it. This is good work.

mcfarhat:
Closing as **superseded by #92**. Adam is no longer working on this.

#92 is `develop` + this branch's beneficiary-widget work + the requested enhancements (friend-name autocomplete with friends-first/Hive fallback, and display/input fit fixes), and it also **folds in the one blocking item from this PR's review** — the deactivation refresh fallback + dropping the premature `notifyDataSetChanged()`. Full credit to @adamdiab2 for the original implementation (carried via Co-authored-by).

Track the feature in #92 going forward.

PR #89, opened 2026-08-13
Add configurable daily step goal

AsmaaZeid:
## What this does
Lets users set their own daily step goal instead of the hardcoded 10,000.

## Changes
- Added a "Daily Step Goal" input card to Settings, matching the existing card style
- Saved/loaded via SharedPreferences (no backend endpoint exists, so local storage is correct per the task spec)
- ChartManager.java now reads the custom goal (via getDailyGoal()) instead of the hardcoded DAILY_GOAL constant — drives both the ring fill percentage and the goal text display

## Verified live on-device
- Setting a custom goal (e.g. 5,000) saves and persists across Settings reopens
- The activity ring correctly displays the custom goal and calculates fill percentage against it

## Open question
The task also mentions this should drive "nudge/milestone messaging." MainActivity.java has separate 5K/7K/10K milestones tied to actual AFIT token reward payouts (RewardManager, claim flow) — deliberately left those untouched, since making them dynamic based on the user's personal goal could let someone claim the 10K reward early by setting a low goal. Flagging this for confirmation — see Trello comment.

## Notes
- Confirmed via gradlew assembleDebug — build succeeds


AsmaaZeid:
PR ready for review: https://github.com/actifit/actifit/pull/89

Core feature done and verified live on-device:
- "Daily Step Goal" setting added, saves/loads via SharedPreferences
- ChartManager.java now uses the custom goal for the ring fill and display
- Confirmed on my phone: setting a custom goal (5,000) persists and correctly changes the ring's fill percentage and goal text

One open question before I call this fully done — the task mentions "nudge/milestone messaging" too. MainActivity.java has separate 5K/7K/10K milestones tied to actual AFIT token rewards (not just display). Making those dynamic based on the personal goal could let someone claim rewards early by setting a low goal, so I left them untouched and only wired the ring display. Let me know if that's the right call or if you meant something else by "milestone messaging."

mcfarhat (changes requested):
Reviewed at `5c691788`. Small, tight, honest diff — and the validation and prefs work is genuinely good. One blocking issue, and I'm sorry to say it's the central claim.

## The ring you fixed isn't the ring on screen

Your PR says *"the activity ring correctly displays the custom goal and calculates fill percentage against it."* The **text** does. The **arc** doesn't. I traced it rather than assume, because it contradicts your device test:

```
ChartManager:155   updateRing(stepRing, ...)              <- your fix, correct
ChartManager:158   ((MainActivity) context).updateDeviceDashboardRings(stepCount)
MainActivity:2871  renderDashboardRings(R.id.dashboard_rings, R.id.step_ring, ...)
MainActivity:2895  private void renderDashboardRings(int auraViewId, int materialRingId, ...)
MainActivity:2916  rings.setActivityRings(steps / 10000f, ...)   <- hardcoded, this is the visible arc
MainActivity:2919  if (materialRing != null) materialRing.setVisibility(View.GONE);   <- hides your ring
```

There are **two overlapping rings** in the same `FrameLayout` (`activity_main.xml:753-823`): the `CircularProgressIndicator step_ring` you corrected, and the `AuraView dashboard_rings` that's `gone` at rest. `renderDashboardRings` makes the AuraView visible and sets `step_ring` to `GONE` — one line after you set its progress. And the AuraView's arc comes from `steps / 10000f`, still hardcoded, clamped 0..1.

`tv_step_goal` is *not* hidden, so your `:143` change does make the centre text read the custom goal.

**What a user sees with a 5,000 goal and 5,000 steps:** the text says the goal is met, and the arc is exactly half full. Same 184dp circle, contradicting itself. At 10,000 steps against a 5,000 goal the text reads 200% while the arc has only just closed.

All three tracking modes route through the same method — device (`:2871`), Fitbit (`:2886`), Health Connect (`:2865`) — so it applies everywhere.

**The fix is one line at `:2916`**, but it needs the goal to be reachable from `MainActivity`. `getDailyGoal()` is currently `private` to `ChartManager`, so lift it to something shared — a `static int getDailyStepGoal(Context)` in `Utils` is the natural home given how many classes need it. While you're there, add `Math.max(1, ...)` around the prefs read: nothing writes 0 today, but that value feeds an integer division and the prefs file is shared by ~50 call sites plus Android auto-backup, so it costs nothing to close the class permanently.

I suspect your device test showed the correct goal *text* and you reasonably read that as the ring working — they're drawn on top of each other. **Please re-test by looking at the arc and the text together** at 5,000 steps with a 5,000 goal.

## Two more sites with the same hardcoded value

- `MainActivity:2946` — `headerAura.setAura(todaySteps / 10000f, ...)`, the small aura around the top-bar avatar. Same "today's goal progress" meaning (`AuraView:153` documents `fillFraction` that way). Decorative, so non-blocking, but it'll visibly disagree with the goal text — fix in the same pass.
- `ProfileActivity:53` + `:220` — its own `DAILY_GOAL = 10000` and `auraStepsFrac = steps / (float) DAILY_GOAL`. The class doc at `:42` says the arc is "today's goal progress". Secondary screen, non-blocking, but list it as follow-up. (It divides by a float, so no zero-division risk there.)

## The judgement call you got right — and it's the best thing in this PR

You left the 5,000 / 7,000 / 10,000 reward milestones hardcoded (`MainActivity:335-337`, `ActivityMonitorService:57`, `HistoryChartActivity:117`, and the share/leaderboard/adapter sites) and said why: making them dynamic would let someone set a 1,000-step goal and claim the 10K AFIT reward early. **That is exactly the exploit**, and those thresholds move real tokens. Raising it in the PR body rather than deciding silently is precisely what I want to see.

One knock-on you should decide about: `ChartManager:128-130` picks the ring colour from `steps >= activityMilestoneOne` (5,000 — a *reward* threshold). With a 3,000 goal the ring reads 100% and stays red. Colour should probably follow `progress >= 100`. Non-blocking, but it affects exactly the users this feature is for.

## Validation — I attacked it and couldn't break it

Credit where it's due. I went looking for a division-by-zero and an escaping `NumberFormatException` and **neither is reachable**:

```
empty        -> caught, old goal kept        0        -> > 0 guard, kept
negative     -> unenterable (inputType)     non-numeric -> unenterable
10^12        -> overflows int, caught, kept  whitespace -> trimmed
decimal      -> unenterable
```

`"dailyStepGoal"` is written only at `SettingsActivity:795-801`, so a 0 can never reach prefs and `(steps * 100) / getDailyGoal()` can't divide by zero. Two residuals, both non-blocking:

1. **Silent rejection.** Clear the field or type `0`, tap Save — everything else commits, the goal quietly doesn't change, no toast, no inline error, and reopening shows the old value. Use `TextInputLayout.setError(...)`.
2. **No upper bound.** `2000000000` parses and stores; progress then reads 0 forever and the goal text will blow out the 11sp `tv_step_goal`. Clamp to something realistic and add `android:maxLength="6"`.

## Prefs — correct, including the bit people get wrong

Same file both sides (`"actifitSets"`, matching all ~50 call sites in the app), consistent `putInt`/`getInt` so no `ClassCastException`, sensible 10,000 default on both reads, and the write reuses the existing editor so it commits with the rest of the save.

And **`getDailyGoal()` is called per `updateRing()` rather than cached in the constructor** — that's the single most common way this feature gets built wrong and you avoided it unprompted.

Worth checking on a device: `MainActivity.onResume` doesn't call `displayActivityChart`, so a goal change may not repaint until the next step-sensor tick. In practice that's soon, but confirm it updates without backgrounding the app.

## Localisation gap

Both new keys are in `values/strings.xml` only — zero occurrences across the 13 translated locales. Graceful (English fallback, not raw keys), but inconsistent with your own practice: `step_goal_format`, the sibling key your `:143` change consumes, **is** translated in all 13, and your #87 adds `retry_action` to all 13. So this looks like an oversight rather than a decision.

**Sequencing note:** #87 already edits those 13 locale files. If you localise here too, a tail-of-file conflict in `values/strings.xml` becomes 13 conflicts. Cheaper to land #87 first, then rebase this. As it stands the two PRs conflict only at the tail of `values/strings.xml` — trivial, but whichever lands second needs a rebase. No functional overlap otherwise, and **this PR does not depend on #87**.

## One thing only you can check

`unofficial_strings.xml` is gitignored and sits in the same `values/` folder. If it already declares `daily_step_goal_lbl` or `steps_suffix` the official build fails on a duplicate resource. `steps_suffix` is generic enough that I'd rate this a real if low risk. I cleared the other four `values/` files; that one I can't see. You said you ran `gradlew assembleDebug` successfully — if that used your local private files, it already answers this, so just confirm it explicitly.

## Layout — your "matches existing card style" claim checks out

I diffed the new card against Card 8 above it attribute by attribute: identical sizing, margins, card background, corner radius, elevation, padding, and title styling. The input block faithfully copies the vote-percent field, same `OutlinedBox` style as all six existing inputs — and you improved on it by putting the suffix in a string resource where the original hardcodes `"%"`. `inputType="number"` present, Material 1.12.0 supports `suffixText`, XML well-formed.

Accessibility: no `labelFor`/`contentDescription`, so TalkBack announces the field as "10000". Worth adding, but `labelFor` appears **zero** times in the whole `res/layout/` tree and the existing vote-percent field has the same gap — a project-wide issue, not yours. Suggestion, not a requirement.

## On verification

The app was **not built** — `google-services.json` and `unofficial_strings.xml` are gitignored and absent, so Gradle can't run. What I did run: `javac 17.0.19` on both changed Java files, giving 868 errors which I classified individually — all `package does not exist` / `cannot find symbol` from the missing Android SDK, and **zero syntax-class errors**. So the files parse cleanly; that's the full extent of the claim. Everything else is a source trace, including my contradiction of your ring result.

Fix the `:2916` line and this is a good, well-scoped feature.

AsmaaZeid:
Fixed. Lifted getDailyGoal() to Utils.getDailyStepGoal(Context) (with the Math.max(1, ...) guard) so it's reachable outside ChartManager, and updated all three hardcoded / 10000f sites to use it: MainActivity:2916 (dashboard ring), MainActivity:2946 (header aura), and ProfileActivity:220 (profile ring).

Retested on a real device this time, watching the arc and text together rather than assuming from the text alone — you were right that I'd been fooled by the two rings overlapping. Confirmed at multiple ratios:

Goal = 5, steps = 5 → ring fully closed, text reads "5/5 steps" 
Goal = 10, steps = 5 → ring ~half filled, text reads "5/10 steps" 
Goal = 20, steps = 5 → ring/text update immediately on returning to the dashboard after saving in Settings, no backgrounding or step-tick needed 

Checked across dashboard, header aura, and profile screen — all three now agree with the text.

Confirming re: unofficial_strings.xml — my gradlew assembleDebug success does use my local private files, so that's not silently masking a duplicate resource risk.

Rebased onto develop now that #87 has merged, to avoid the strings.xml tail conflict you flagged — clean rebase, one expected conflict on that file, resolved by keeping both sets of strings.

Left as non-blocking follow-ups per your notes: ring colour still tied to the fixed 5,000-token milestone rather than progress >= 100, no upper bound on goal input, and silent rejection on empty/0 input. Happy to take these in a follow-up if useful.

mcfarhat (approved):
## Reviewed at `0d32f6e1` — approving and merging.

The follow-up commits do what I asked. `getDailyGoal()` is lifted to `Utils.getDailyStepGoal(Context)` as a single source of truth, and every consumer now goes through it.

### The guard is in the right place

```java
int goal = prefs.getInt("dailyStepGoal", 10000);
return Math.max(1, goal);
```

That matters because three of the four call sites divide by it — `(steps * 100) / getDailyGoal()` in `ChartManager.updateRing`, and the two `steps / (float) …` ring fractions. Putting the clamp in the getter rather than at each call site means a `0` in SharedPreferences (from an older build, a bad write, whatever) can't reach a divisor. Defending it once at the source is the right shape.

I traced all four consumers and they're consistent:

```
ChartManager:57       ring progress + goal label
MainActivity:2916     dashboard activity rings
MainActivity:2946     header aura
ProfileActivity:220   profile activity ring
```

### The trap you didn't fall into

The card says the goal drives "the activity-ring fill + the nudge/milestone messaging", which reads like an instruction to make `activityMilestoneOne/Two/Three` (5000/7000/10000) follow the setting too. **You correctly left them alone.**

Those aren't personal goals — they're AFIT reward thresholds. They're passed to `RewardManager` at `MainActivity:859` and annotated at 1658–1678 as "Steps needed for 5k / 7k / 10k reward". Wiring them to a user-editable preference would let anyone set a goal of 100 and claim the 10k reward. Keeping the ring colour at `steps >= activityMilestoneOne` while the fill follows the configured goal is exactly the right split: personal target drives the display, protocol thresholds drive the payout.

### Save path

Defensive, and I checked the edge cases rather than assuming:

```java
try {
    int stepGoal = Integer.parseInt(dailyStepGoalInput.getText().toString().trim());
    if (stepGoal > 0) editor.putInt("dailyStepGoal", stepGoal);
} catch (NumberFormatException e) { }
```

Empty field, non-numeric, and values that overflow `int` all land in the catch and leave the stored goal untouched; `android:inputType="number"` blocks a leading minus. Nothing can persist a zero or negative.

### One thing worth knowing — the ring doesn't refresh until relaunch

`displayActivityChart` is what owns both the ring fill and the goal label. I traced every call site of it in `MainActivity`:

```
1829, 2225   onCreate
4289, 4321   PrepareGround.onPostExecute  (executed from onCreate:1424)
TrackingManager:451  useDefaultTrackingMethod  (permission/tracking init only)
```

`onResume` (3890) runs `DisplayDayChartDataAsyncTask` and `DisplayChartDataAsyncTask` — the day/history charts — but never `displayActivityChart`. And `SettingsActivity` doesn't `recreate()` or relaunch `MainActivity` on save.

So the sequence *change goal → back → dashboard* leaves the ring fill and the "goal" text showing the previous value until the activity is recreated (app relaunch, rotation, process death) or tracking re-initialises. The value is stored correctly and every consumer reads it correctly — this is purely a redraw trigger.

Not blocking the merge: it's a one-line `displayActivityChart(...)` in `onResume`, it self-corrects on next launch, and no data is wrong at any point. But it is the feature's headline moment, so it's worth a follow-up card.

**Caveat on that finding:** there's no Android SDK on my machine, so I could not build or run this. The above is call-graph tracing, not observed behaviour — if there's a step-sensor broadcast path that redraws the ring which I didn't find, this doesn't apply. Everything else in this review is verified by reading the code at this head.

### Minor leftovers

- `ProfileActivity:53` — `private static final int DAILY_GOAL = 10000;` is now dead; the only reference was the line you replaced at 220.
- `ChartManager:39` — `import android.content.SharedPreferences;` was added but nothing in that file uses it, since you delegate to `Utils`.
- Invalid or cleared input is silently ignored on save. The field is pre-populated, so clearing it is a plausible user action and they get no feedback that the change didn't take.

### Resource check

In place of a compile, I resolved every resource the change references: `daily_step_goal_lbl`, `steps_suffix`, `step_goal_format`, `step_pct_to_goal_format` all exist, `R.id.dailyStepGoalInput` matches the layout, and `metric_system_ntt` / `us_system_ntt` resolve out of `non_translatable_strings.xml` (those two lines only moved during the javadoc commit).

Merging. Clean feature, and the reward-threshold distinction was the judgement call that mattered here.

